PR #686, opened 2023-01-24
fix/feat: 891 reset all collect record input/observations on input ch…

mmnoo:
Description:

when a user changes a collect record input or observation row input with any type of validation, the validation info clears, and the validation gets marked as 'reset' in browser storage (to update the api, which doesnt have its own logic for this yet)


mmnoo:
[Front end: reset/remove input/observation validations on any input change](https://trello.com/c/x7vecZb1/891-front-end-reset-remove-input-observation-validations-on-any-input-change)

PR #734, opened 2023-02-16
fix: [M957] Fix default mr rule input.

tienhoah:


tienhoah:
[MR default rule not saving on add](https://trello.com/c/jIjc5v0y/957-mr-default-rule-not-saving-on-add)

mmnoo (approved):

PR #742, opened 2023-03-01
Melissa/842 3

mmnoo:
A speculative improvements for a bug that I couldnt reproduce (see ticket comments https://trello.com/c/k4O2MjrT/842-audit-cryptic-toast-error-messages)
speeding up pre push test hook

### DIFF
--- a/package.json
+++ b/package.json
@@ -107,7 +107,7 @@
   "husky": {
     "hooks": {
       "pre-commit": "yarn run prettier",
-      "pre-push": "yarn run lint && yarn run test --watchAll=false"
+      "pre-push": "yarn run lint && yarn run test --watchAll=false --silent"
     }
   },
   "jest": {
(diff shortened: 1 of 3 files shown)


mmnoo:
[audit cryptic toast error messages](https://trello.com/c/k4O2MjrT/842-audit-cryptic-toast-error-messages)

julia-rom (commented):
had a look at the ticket and a bit confused - is the issue here that the error is unclear? my assumption would be that it is an api error and to try refreshing. How will this make it clearer to the user?

mmnoo:
> had a look at the ticket and a bit confused - is the issue here that the error is unclear? my assumption would be that it is an api error and to try refreshing. How will this make it clearer to the user?

good question. This bug came up as part of the audit work, but I dont have enough details on the bug itself, but we discussed verbally that I reverse engineer what might possibly be triggering it and fix it, even if speculatively. So I added some try/catch and .?

julia-rom:
ah got it - as long as that's flagged in the ticket i'm ok to approve!

PR #745, opened 2023-03-06
account for longer collect count

alanjleonard:


alanjleonard:
[The circle that shows the number of sample units in collecting next to the collecting tab only fits two characters, so the third is placed under it.](https://trello.com/c/jnE8CV30/936-the-circle-that-shows-the-number-of-sample-units-in-collecting-next-to-the-collecting-tab-only-fits-two-characters-so-the-third)

mmnoo (approved):

PR #746, opened 2023-03-07
Melissa/968 test speed

mmnoo:
Ticket: https://trello.com/c/tgDctEzq/968-research-making-tests-faster-timebox-4-hours

This change involves two steps. 
The first is changing how state is fed into the app to allow easier mocking and rendering of isolated component free of them being wrapped in the app state (which is mocked by test renderer functions instead)
The second involves selecting a sample of integration tests and replacing them rendering the whole app, and rendering the page component instead in an attempt to speed up the test suite

Results running the sample tests all together:
This branch:         
11.6 seconds average
Develop:
21.7 seconds average

Thats almost twice as fast!

However the tradeoffs are that we are relying more on mocks to make tests run. This isnt great ideally, but its better to have tests that are twice as fast which makes them more useful and also able to run in CI.

Another tradeoff that is more meaningful IMO is that writing tests this new way and relying on the render mocks more makes the test suite even more complicated to understand and maintain. Check out some of the new tests and let me know what you think. Maybe its the same complexity/confusingness? 

Note: I have broken some of the untouched tests with this change, but am pushing this PR for conversation purposes. If you want to run the sample tests use `yarn test src/App/integrationTests/collectRecords/habitatComplexity/App.updateHabitatComplexity.test.js src/App/integrationTests/collectRecords/habitatComplexity/App.validationDisplay1HabitatComplexity.test.js src/App/integrationTests/collectRecords/habitatComplexity/App.validationDisplay2HabitatComplexity.test.js src/App/integrationTests/collectRecords/habitatComplexity/App.validationIgnoring1HabitatComplexity.test.js src/App/integrationTests/collectRecords/habitatComplexity/App.validationIgnoring2HabitatComplexity.test.js `

### DIFF
--- a/src/App/App.js
+++ b/src/App/App.js
@@ -14,7 +14,6 @@ import { useOnlineStatus } from '../library/onlineStatusContext'
 import { useRoutes } from './useRoutes'
 import { useSyncStatus } from './mermaidData/syncApiDataIntoOfflineStorage/SyncStatusContext'
 import DatabaseSwitchboard from './mermaidData/databaseSwitchboard'
-import dexieInstancePropTypes from './dexieInstancePropTypes'
 import Footer from '../components/Footer'
 import GlobalStyle from '../library/styling/globalStyles'
 import Header from '../components/Header'
@@ -28,12 +27,14 @@ import useIsMounted from '../library/useIsMounted'
 import { useDexiePerUserDataInstance } from './dexiePerUserDataInstanceContext'
 import handleHttpResponseError from '../library/handleHttpResponseError'
 import ErrorBoundary from '../components/ErrorBoundary'
+import { useDexieCurrentUserInstance } from './dexieCurrentUserInstanceContext'
 
-function App({ dexieCurrentUserInstance }) {
+function App() {
   const { isAppOnline } = useOnlineStatus()
   const apiBaseUrl = process.env.REACT_APP_MERMAID_API
   const isMounted = useIsMounted()
   const { isOfflineStorageHydrated, syncErrors, isSyncInProgress } = useSyncStatus()
+  const dexieCurrentUserInstance = useDexieCurrentUserInstance()
 
   const { getAccessToken, isMermaidAuthenticated, logoutMermaid } = useAuthentication({
     dexieCurrentUserInstance,
@@ -188,10 +189,6 @@ function App({ dexieCurrentUserInstance }) {
       </DatabaseSwitchboardInstanceProvider>
     </ThemeProvider>
   )
-}
-
-App.propTypes = {
-  dexieCurrentUserInstance: dexieInstancePropTypes.isRequired,
 }
 
 export default App
--- a/src/App/dexieCurrentUserInstanceContext.js
+++ b/src/App/dexieCurrentUserInstanceContext.js
@@ -1,0 +1,41 @@
+import PropTypes from 'prop-types'
+import React, { createContext, useContext, useState } from 'react'
+
+import dexieInstancePropTypes from './dexieInstancePropTypes'
+import dexieCurrentUserInstance from './dexieCurrentUserInstance'
+
+const DexieCurrentUserInstanceContext = createContext()
+
+const DexieCurrentUserInstanceProvider = ({ children, value }) => {
+  const isTestEnvironment = process.env.NODE_ENV === 'test'
+  const currentUserInstanceToUse = isTestEnvironment ? value : dexieCurrentUserInstance
+
+  return (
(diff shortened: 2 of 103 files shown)


mmnoo:
[research making tests faster (timebox ~4 hours)](https://trello.com/c/tgDctEzq/968-research-making-tests-faster-timebox-4-hours)

JBurkinshaw:
> However the tradeoffs are that we are relying more on mocks to make tests run. This isnt great ideally, but its better to have tests that are twice as fast which makes them more useful and also able to run in CI.

I think this is fine if it prevents the same process being repeated every time a test runs. No need to keep running the same thing in the tests if we we already know it works.

> Another tradeoff that is more meaningful IMO is that writing tests this new way and relying on the render mocks more makes the test suite even more complicated to understand and maintain. Check out some of the new tests and let me know what you think. Maybe its the same complexity/confusingness?

The added complexity is definitely not ideal but if the tradeoff is that it makes tests less likely to randomly or mysteriously fail in a way that is difficult to debug, I think it is worth it. I looked at `App.updateHabitatComplexity.test.js` and I don't thing the increase in complexity is too drastic.

mmnoo:
> > However the tradeoffs are that we are relying more on mocks to make tests run. This isnt great ideally, but its better to have tests that are twice as fast which makes them more useful and also able to run in CI.
> 
> I think this is fine if it prevents the same process being repeated every time a test runs. No need to keep running the same thing in the tests if we we already know it works.
> 
> > Another tradeoff that is more meaningful IMO is that writing tests this new way and relying on the render mocks more makes the test suite even more complicated to understand and maintain. Check out some of the new tests and let me know what you think. Maybe its the same complexity/confusingness?
> 
> The added complexity is definitely not ideal but if the tradeoff is that it makes tests less likely to randomly or mysteriously fail in a way that is difficult to debug, I think it is worth it. I looked at `App.updateHabitatComplexity.test.js` and I don't thing the increase in complexity is too drastic.

@JBurkinshaw this only fixes speed, not the random failing. The main change is that we are rendering just the component in question, not the entire app and its dom tree. If anything, relying more on mocking will increase 'random' errors and increase false passes as well (because we are testing the mocks, not the app - they become less of integration tests and more like unit tests)

mmnoo:
The more I think about this approach, the more I dont like it as its likely to complicate dev onboarding and test troubleshooting (and lose the value of leaning more toward an integration test instead of the new approach relying so much on mocking for app state)

@saanobhaai if we were to cut the test run speed by 80%, how much money would we be saving form GH actions, and is it worth the extra complicated-ness and risk to requiring extra dev time to deal with slightly more finicky and less useful (in theory) tests?

On the dev front cutting the tests in half means still reasonably slow test suite, so for local dev process only, I dont think its worth it to make the change.

julia-rom:
That's awesome that you were able to speed it up significantly @mmnoo ! I do agree that going in these tests can feel complicated for a dev to wrap their head around easily. I don't, however, feel like the changes make it significantly more complicated. Main thing is we are testing for a bunch of things in each test. I know this helps speed things up, but it also makes it harder to understand what's going on exactly in each test.

I think we should be doing more unit tests on components, and figuring out ways to separate integration tests. Not all component tests need to be integration and vice versa. I don't have the answer of that will look like for each component, but it could be worth putting more time together to figure out what can be a unit test (and just use a mock), and where we feel mocks are not appropriate (and just render the whole app)

mmnoo:
@julia-rom what do you think of this article on unit vs integration tests? https://kentcdodds.com/blog/write-tests

Its informed my approach to testing in the last few years. I dont feel very assured when unit tests pass. 

https://twitter.com/erinfranmc/status/1148986961207730176

julia-rom:
ahh love Kent Dodds, actually a good reminder to re-read some of his stuff. I like his focus on integration tests, as the point about 

> It doesn't matter if your component <A /> renders component <B /> with props c and d if component <B /> actually breaks if prop e is not supplied. 

I agree with the focus on integration tests, i'm more so just thinking where we can split them apart more (I don't have a specific answer for this yet). 

also the umbrella gif is gold 😆... I feel like i'm going to go into a bit of a kent dodds wormhole right now because its been a while since I read his stuff

mmnoo:
Totally! I idolize very few dev celebrities, but KCD is consistently have very good content. I really like how non macho he is about tech, and he communicates very very well.

mmnoo:
I used to do mostly unit tests and then of course when I started working with React, I bumped into his content, and it totally changed my approach to testing. I think he is one of the main contributors to Testing Library.

mmnoo:
Closing this to clear the queue. We can discuss further later if needed.

PR #747, opened 2023-03-07
Nick/M966 fix mr duplicate bugs

tienhoah:
fix mr duplicate bugs
fix unique key names in Observers and Transects table

### DIFF
--- a/src/components/SampleUnitPopups/CollectSampleUnitPopup/CollectSampleUnitPopup.js
+++ b/src/components/SampleUnitPopups/CollectSampleUnitPopup/CollectSampleUnitPopup.js
@@ -18,12 +18,13 @@ const CollectSampleUnitPopup = ({ rowRecord, recordProfileSummary }) => {
 
     const popupTitle = `${sample_unit_method} ${name}`
     const transectNumberLabel = name || language.pages.usersAndTransectsTable.missingLabelNumber
-    const keyName = transectNumberLabel + index
 
     const managementName =
       management_name === API_NULL_NAME
         ? language.pages.usersAndTransectsTable.missingMRName
         : management_name
+
+    const keyName = transectNumberLabel + site_name + managementName + profile_name + sample_date
 
     return (
       <SampleUnitNumber tabIndex="0" id={index} key={keyName}>
--- a/src/components/SampleUnitPopups/CollectSampleUnitPopup/CollectSampleUnitPopup.js
+++ b/src/components/SampleUnitPopups/CollectSampleUnitPopup/CollectSampleUnitPopup.js
@@ -15,16 +15,18 @@ const CollectSampleUnitPopup = ({ rowRecord, recordProfileSummary }) => {
 
   const sampleUnitsWithPopup = sortArray(collect_records).map((record, index) => {
     const { name, sample_date, observers, management_name } = record
-    const popupTitle = `${sample_unit_method} ${name}`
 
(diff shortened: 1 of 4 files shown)


src/components/SampleUnitPopups/CollectSampleUnitPopup/CollectSampleUnitPopup.js
```diff
@@ -15,16 +15,18 @@ const CollectSampleUnitPopup = ({ rowRecord, recordProfileSummary }) => {
 
   const sampleUnitsWithPopup = sortArray(collect_records).map((record, index) => {
     const { name, sample_date, observers, management_name } = record
-    const popupTitle = `${sample_unit_method} ${name}`
 
+    const popupTitle = `${sample_unit_method} ${name}`
     const transectNumberLabel = name || language.pages.usersAndTransectsTable.missingLabelNumber
+    const keyName = transectNumberLabel + index
```
mmnoo, line 21: is there a unique id that can be used from `record` instead of `index`? (React doesnt like you using index for the key value)

src/components/SampleUnitPopups/SubmittedSampleUnitPopup/SubmittedSampleUnitPopup.js
```diff
@@ -20,9 +20,10 @@ const SubmittedSampleUnitPopup = ({ rowRecord, sampleUnitNumbersRow }) => {
       management.name === API_NULL_NAME
         ? language.pages.usersAndTransectsTable.missingMRName
         : management.name
+    const keyName = transectNumberLabel + index
```
mmnoo, line 23: same here

src/components/SampleUnitPopups/CollectSampleUnitPopup/CollectSampleUnitPopup.js
tienhoah: no unique id for them. I have some options

1) I bypass the eslint error by doing as above
2) I have to make a very long string name with different properties (its not guaranteed they are unique)
3) Use random id generations (is it legal >.<?) 

Any thoughts?

mmnoo: 3 is illegal
Im guessing it complains about `transectNumberLabel` not being unique.... because they are sometimes undefined (and use the same languege file text)

mmnoo: In that case just do what you were doing. Its silly they wont let you silence the warning explicitly.

mmnoo (approved):

src/components/SampleUnitPopups/CollectSampleUnitPopup/CollectSampleUnitPopup.js
tienhoah: okay, I change to option 2 for now, it might be very edge case to get the same key names.

tienhoah: yeah it couldn't tell. thanks I will merge with option 2

tienhoah:
[resolve MR duplicate Parties errors](https://trello.com/c/vkTy7SmI/966-resolve-mr-duplicate-parties-errors)

PR #749, opened 2023-03-08
fix: M971 misleading site save messaging when offline

mmnoo:
Steps to test:
- create a new site while offline. 
- click save

Expected results
- a toast with "The site has been saved on your computer." shows

Do the same when online: Expect the toast to say "The site has been saved on your computer and in the MERMAID online system."

Repeat (online and off) for saving an MR

### DIFF
--- a/src/App/integrationTests/managementRegime/App.managementRegimeCreateOffline.test.js
+++ b/src/App/integrationTests/managementRegime/App.managementRegimeCreateOffline.test.js
@@ -87,11 +87,7 @@ describe('Offline', () => {
 
     await saveMR()
 
-    expect(
-      await screen.findByText(
-        'The management regime has been saved on your computer and in the MERMAID online system.',
-      ),
-    )
+    expect(await screen.findByText('The management regime has been saved on your computer.'))
 
     // ensure the new form is now the edit form
     expect(await screen.findByTestId('edit-management-regime-form-title'))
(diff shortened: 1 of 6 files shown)


mmnoo:
[Site saved success message misleading when offline](https://trello.com/c/tsey52XF/971-site-saved-success-message-misleading-when-offline)

src/App/integrationTests/managementRegime/App.managementRegimeCreateOffline.test.js
```diff
@@ -87,11 +87,7 @@ describe('Offline', () => {
 
     await saveMR()
 
-    expect(
-      await screen.findByText(
-        'The management regime has been saved on your computer and in the MERMAID online system.',
-      ),
-    )
+    expect(await screen.findByText('The management regime has been saved on your computer.'))
```
julia-rom, line 90: is it worth also mentioning that it will attempt to save on the mermaid db once tehre's a steady internet connection?

julia-rom (approved):
its ok - I feel like I would improve the text a bit by clarifying when it hasn't been added to the mermaid online system yet + will try to do so once we have an active/steady internet connection

src/App/integrationTests/managementRegime/App.managementRegimeCreateOffline.test.js
mmnoo: Probably, but Ill leave wording to @alanjleonard. He has a task to look through error messages and improve language as needed.

(I like that idea FWIW)

PR #754, opened 2023-03-09
fix: 975 data sharing page. Added missing info for read only mode.

mmnoo:
To test:

- Login to a project that you are admin on
- Go to the data sharing page, toggle the 'this is a test project' checkmark to be checked
- log out, 
- log back in as a user that is read only or collector
- view the data sharing page

Expected results: 
- you should see some text that says 'This is a test project' with no checkbox input (because we are read only now)

Other test:
- log back in as admin
- go to datasharing page
- toggle the 'This is a test project' checkbox to be unchecked
- log out and back in as a read only or collector
- view the data sharing page

Expected results: 
- you should NOT see text that says 'This is a test project' (and there should also be no checkbox for it)


mmnoo:
[Missing project fields in read-only mode](https://trello.com/c/MwInT1oD/975-missing-project-fields-in-read-only-mode)

julia-rom (approved):
code looks good, tests pass!

PR #759, opened 2023-03-13
Nick/M955 - Creating tests for duplicate (Site & MR) modals for all sample units

tienhoah:


src/App/integrationTests/collectRecords/benthicLit/App.duplicateModal.test.js
```diff
@@ -0,0 +1,354 @@
...
+  })
+
+  userEvent.click(editOriginalSite)
+
+  await waitForElementToBeRemoved(() => screen.queryByLabelText('project pages loading indicator'))
+  // const siteCPage = await screen.findByText('Site C', {
+  //   selector: 'h2',
+  // })
+
+  // expect(siteCPage).toBeInTheDocument()
```
tienhoah, line 353: this part fails

tienhoah:
[Write new feature tests for 917, 916](https://trello.com/c/IdkuyXre/955-write-new-feature-tests-for-917-916)

src/App/integrationTests/collectRecords/benthicLit/App.validationDuplicateModalBenthicLit.test.js
```diff
@@ -0,0 +1,718 @@
...
+      initialEntries: ['/projects/5/collecting/benthiclit/70'],
+    },
+    dexiePerUserDataInstance,
+    dexieCurrentUserInstance,
+  )
+
+  await validateCollectRecordAndResolveSimilarSite()
+
+  const resolveModal = screen.getByTestId('resolve-duplicate-site')
+  const duplicateSite = await within(resolveModal).findByTestId('duplicate-site')
```
mmnoo, line 333: I think for these headings we can grab the element by other means. Try `.getByRole('columnheader', {name 'Duplicate site})`

src/components/ResolveDuplicateMRButtonAndModal/ResolveDuplicateMRButtonAndModal.js
```diff
@@ -81,13 +86,19 @@ const ResolveDuplicateMRButtonAndModal = ({
...
+                  getOptions(choicesResponse.managementcompliances.data),
+                )
+                setCurrentManagementRegimeData(currentManagementRegimeResponse)
+                setDuplicateManagementRegimeData(duplicateManagementRegimeResponse)
+                setIsResolveDuplicateModalLoading(false)
+              }
             },
           )
           .catch((error) => {
+            setIsResolveDuplicateModalLoading(false)
```
mmnoo, line 101: Sweet! You caught a bug!

src/components/ResolveDuplicateSiteButtonAndModal/ResolveDuplicateSiteButtonAndModal.js
```diff
@@ -312,10 +334,19 @@ const ResolveDuplicateSiteButtonAndModal = ({
...
+        testId="resolve-duplicate-site"
         title="Resolve Duplicate"
         isOpen={isResolveDuplicateModalOpen}
         onDismiss={closeResolveDuplicateModal}
-        mainContent={mainContent}
+        mainContent={
+          isResolveDuplicateModalLoading ? (
+            <ModalLoadingIndicatorWrapper>
+              <LoadingIndicator />
+            </ModalLoadingIndicatorWrapper>
```
mmnoo, line 345: Its always loaders, isnt it?

mmnoo (commented):
Overall this looks fantastic. I like how through it is. We can see just how complex the resolving duplicates feature is and the various scenarios that are involved in making it work.

Two things that would make it better are
- removing assertions from your `validateCollectRecordAndResolveSimilarSite` utility. All expect statements shoudl be in the body of the test itself.
- avoiding as many of the uses of data-testid that are possible to avoid. Ive put some suggestions in.

src/testUtilities/mockCollectRecords/mockBenthicLitValidationsObject.js
```diff
@@ -1,4 +1,4 @@
-export const mockBenthicLitValidationsObject = {
+export default {
```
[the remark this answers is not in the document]
tienhoah: I see, I'm making this consistent for all mock data.

src/components/ResolveDuplicateSiteButtonAndModal/ResolveDuplicateSiteButtonAndModal.js
tienhoah: yeah, its modal loading created by AL used in Copy Sites and MRs modals.

mmnoo: I mean, a lot of testing mishaps are often due to a lack of loaders or not waiting for loaders. I assumed this change was that.

mmnoo:
I commented, but really I will look at this again tomorrow

tienhoah:
> I commented, but really I will look at this again tomorrow

I made changes, thanks for the comments

src/components/ResolveDuplicateMRButtonAndModal/ResolveDuplicateMRButtonAndModal.js
```diff
@@ -243,11 +262,9 @@ const ResolveDuplicateMRButtonAndModal = ({
                 <IconPen /> {editEither}
               </ButtonCaution>
             </Thead>
-            <Thead>
+            <Thead aria-label="Duplicate Management">
```
mmnoo, line 265: This is good enough for now, but in future tests, we should try to use selectors like `getByRole('columnheader', {name: 'bla bla bla'})`. Here this aria-label is unnecessary (to my knowledge) for a screen reader because the screen reader can tell its a column header (th) and what the text within is. In general using aria-label to help with tests is better than using test-id, but using other selectors is better than either of those unless aria label actually helps a screen reader that has no other way of interpreting the markup. 

Often in those cases, if time allows, its good to talk to Al about it, because those times are usually pointing out where the UX could use some explicit messaging (eg icon buttons)

But lets move on. We can work on that in the next set of tests. 

This is very good for being newish to authoring completely from scratch a bunch of tests in this very complicated application

mmnoo (approved):

src/components/ResolveDuplicateMRButtonAndModal/ResolveDuplicateMRButtonAndModal.js
mmnoo: ah, ya, then my advice is cheap. Ill try pulling it and seeing. Feel free to ignore and merge if you would rather.

mmnoo: K, so I figured out waht the issue is with selecting the table headers. 

This will work:
```
await within(resolveModal).findByRole('columnheader', {
    name: 'foo',
  })
  ```
  
  for 
  `<th scope="col">foo </th>` or `<th role="columnheader">foo</th>`
  
  
  But it will not work if there are other things nested in the `th` like buttons, which is what is going on with the resolve duplicates table.
  
  ```
  <th scope="col">
              foo <button>other stuff</button>
    </th>
            ```
            
            So `aria-label` is the right call for the ths that also have other stuff in them :)

PR #760, opened 2023-03-13
fix: [M976] Change popover date label.

tienhoah:


tienhoah:
[change Observers and Transects popover Date label](https://trello.com/c/cvmSi2lS/976-change-observers-and-transects-popover-date-label)

mmnoo (approved):

PR #763, opened 2023-03-14
Nick/M804 - Link fish size bin and fish observations

tienhoah:
Fish bin size and observation table should be linked. 
- Shouldn’t be able to change the fish bin size if there’s observations.
- Shouldn’t be able to add any observations if fish bin size isn’t selected.

### DIFF
--- a/src/components/mermaidInputs/InputValidationInfo/InputValidationInfo.js
+++ b/src/components/mermaidInputs/InputValidationInfo/InputValidationInfo.js
@@ -23,6 +23,7 @@ const InputValidationInfo = ({
   validationType,
   currentSelectValue,
   updateValueAndResetValidationForDuplicateWarning,
+  hasObservationRecords,
 }) => {
   const { projectId } = useParams()
   const areThereValidationMessages = validationMessages.length
@@ -93,6 +94,7 @@ const InputValidationInfo = ({
           ))}
         </>
       ) : null}
+      {hasObservationRecords ? <>{language.error.disabledFishSizeBinSelect}</> : null}
       {isWarningValidation || isIgnoredWarningValidation ? getWarningValidationButtons() : null}
       {isValidationPassing ? <span aria-label="Passed Validation">&nbsp;</span> : null}
     </ValidationWrapper>
@@ -106,6 +108,7 @@ InputValidationInfo.propTypes = {
   validationMessages: mermaidInputsPropTypes.validationMessagesPropType,
   currentSelectValue: PropTypes.string,
   updateValueAndResetValidationForDuplicateWarning: PropTypes.func,
+  hasObservationRecords: PropTypes.bool,
 }
 
 InputValidationInfo.defaultProps = {
@@ -115,6 +118,7 @@ InputValidationInfo.defaultProps = {
   validationType: undefined,
   currentSelectValue: undefined,
   updateValueAndResetValidationForDuplicateWarning: () => {},
+  hasObservationRecords: false,
 }
 
 export default InputValidationInfo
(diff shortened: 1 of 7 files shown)


tienhoah:
[Link fish size bin and observations table](https://trello.com/c/DIHPNTjM/804-link-fish-size-bin-and-observations-table)

src/components/mermaidInputs/InputValidationInfo/InputValidationInfo.js
```diff
@@ -23,6 +23,7 @@ const InputValidationInfo = ({
   validationType,
   currentSelectValue,
   updateValueAndResetValidationForDuplicateWarning,
+  hasObservationRecords,
```
mmnoo, line 26: I think this is too specific to a fish bin input. This component is a generic input component (generic to Mermaid anyway) Maybe the prop be more generic like `additionalText` and then below so `additionalText ? additionalText : null` and let the parent decide when to include the additional text. 

this way it is potentially resulable for other components that need non-validation helper text (but dont share the rule about observation tables)

src/components/pages/collectRecordFormPages/FishBeltForm/FishBeltTransectInputs.js
```diff
@@ -89,6 +93,10 @@ const FishBeltTransectInputs = ({
     areValidationsShowing,
   )
 
+  // const sizeBinValidationInfo = hasSizeBinValidation
```
mmnoo, line 96: dont forget to remove commented code :)

src/components/pages/collectRecordFormPages/FishBeltForm/FishBeltTransectInputs.js
```diff
@@ -358,6 +367,7 @@ const FishBeltTransectInputs = ({
             sizeBinValidationProperties,
             'size_bin',
           )}
+          hasObservationRecords={hasFishBeltObservations}
           value={formik.values.size_bin}
```
mmnoo, line 371: you could make this something like `additionalText={hasFishbeltObservations ? language.whatever.whatever : null}`

You can think of this suggestion as inverting control logic to the parent instead of the child.

mmnoo (changes requested):
I really like the ux of this. I tested out the preview app, and it looks good!

One request is to keep the input component simple and generic and move the logic up to the parent/composing component.  Other than that, looks good.

tienhoah:
@mmnoo Thanks for great suggestion, I made a similar adjustment as your approach

src/components/pages/collectRecordFormPages/FishBeltForm/FishBeltObservationTable.js
```diff
@@ -430,9 +432,16 @@ const FishBeltObservationTable = ({
...
-        <ButtonPrimary type="button" onClick={handleAddObservation}>
-          <IconPlus /> Add Row
-        </ButtonPrimary>
+        <UnderTableRowButtonArea>
+          <ButtonPrimary
+            type="button"
+            onClick={handleAddObservation}
+            disabled={!fishBinSelectedLabel}
+          >
+            <IconPlus /> Add Row
```
mmnoo, line 441: Not suggesting you change this here since pushing more changes and retesting is a pain, but as we work through code moving forward, and we bump into old hard coded text, can we start adding it to the language file?

mmnoo (approved):
FAntastic. Sorry I missed your email to rereview this and it sat for longer than it should have.

src/components/pages/collectRecordFormPages/FishBeltForm/FishBeltObservationTable.js
tienhoah: noted

PR #765, opened 2023-03-16
draft: ideas for collect record title/ 'sub nav node title' dont refresh when inputs updated and saved

julia-rom:
turned your 'ideas branch'  into a draft so I can comment - no need to actually review @mmnoo

### DIFF
--- a/src/components/pages/collectRecordFormPages/HabitatComplexityForm/HabitatComplexityForm.js
+++ b/src/components/pages/collectRecordFormPages/HabitatComplexityForm/HabitatComplexityForm.js
@@ -114,8 +114,12 @@ const HabitatComplexityForm = ({ isNewRecord }) => {
     )
   }, [collectRecordBeingEdited, getPersistedUnsavedFormikData])
 
-  const handleCollectRecordChange = (updatedCollectRecord) =>
+  const handleCollectRecordChange = (updatedCollectRecord) => {
+    // maybe rename this function. Its pretty vague...
+    // need to recompute record name for sub nav node for all CRs
+    // setSubNavNode(recordNameForSubNode)
     setCollectRecordBeingEdited(updatedCollectRecord)
+  }
 
   return (
     <ErrorBoundary>
--- a/src/components/pages/collectRecordFormPages/HabitatComplexityForm/HabitatComplexityForm.js
+++ b/src/components/pages/collectRecordFormPages/HabitatComplexityForm/HabitatComplexityForm.js
@@ -26,12 +26,20 @@ import language from '../../../../language'
 import useIsMounted from '../../../../library/useIsMounted'
 import ErrorBoundary from '../../../ErrorBoundary'
 
+// I wonder if we can make this reusable for all protocols
+
+const getDataForSubNavNode = ({ isNewRecord, collectRecord, sites }) =>
+  !isNewRecord && collectRecord
+    ? getRecordSubNavNodeInfo(collectRecord.data, sites, collectRecord.data.protocol)
+    : { name: language.protocolTitles.habitatcomplexity }
+
 const HabitatComplexityForm = ({ isNewRecord }) => {
   const [areObservationsInputsDirty, setAreObservationsInputsDirty] = useState(false)
   const [collectRecordBeingEdited, setCollectRecordBeingEdited] = useState()
   const [idsNotAssociatedWithData, setIdsNotAssociatedWithData] = useState([])
   const [isLoading, setIsLoading] = useState(true)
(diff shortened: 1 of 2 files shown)


src/components/pages/collectRecordFormPages/HabitatComplexityForm/HabitatComplexityForm.js
```diff
@@ -114,8 +114,12 @@ const HabitatComplexityForm = ({ isNewRecord }) => {
     )
   }, [collectRecordBeingEdited, getPersistedUnsavedFormikData])
 
-  const handleCollectRecordChange = (updatedCollectRecord) =>
+  const handleCollectRecordChange = (updatedCollectRecord) => {
+    // maybe rename this function. Its pretty vague...
+    // need to recompute record name for sub nav node for all CRs
+    // setSubNavNode(recordNameForSubNode)
```
julia-rom, line 120: in the `updatedCollectRecord` response, we have  `data.sample_event.site` with a long id string, I *pretty sure* I remember seeing a function that decodes this into a name, but don't remember which one @mmnoo

julia-rom: is the the site id ? why don't we also include the actual names in the response

mmnoo: That is a good question. I dont know. It would make the front end code cheaper to build/maintain. And probably faster (avoiding reading IDB extra times to join data)..

I think the goal at the time was to change the api minimally from v1... and it also had to be backward compatible. Now that we are killing v1, maybe there are opportunities moving forward for new features at least.

mmnoo: yes its site id

mmnoo: But that setSubNode function should do it for you magically if make this code above reusable here

```
              const recordNameForSubNode =
                !isNewRecord && collectRecordResponse
                  ? getRecordSubNavNodeInfo(
                      collectRecordResponse.data,
                      sitesResponse,
                      collectRecordResponse.data.protocol,
                    )
                  : { name: language.protocolTitles.habitatcomplexity }
                  ```

Comment (julia-rom) [reply src/components/pages/collectRecordFormPages/HabitatComplexityForm/HabitatComplexityForm.js:120]: @mmnoo so setSubNode uses `getRecordSubNavNodeInfo` which requires 3 params -` collectRecordResponse`, `sites`, and `protocol`.  

we get `sites` from  `databaseSwitchboardInstance.getSitesWithoutOfflineDeleted` . it looks like we would have to call this again within `handleCollectRecordChange` to get fresh sites data incase one of the names changes, however, we would have to somehow check if that was the case. in this case,  it feels like we are doing too much just to get the site name, which is included in the `sample_event` info, however its as an id string

Comment (julia-rom) [reply src/components/pages/collectRecordFormPages/HabitatComplexityForm/HabitatComplexityForm.js:120]: I think now that we are killing v1, it would be good to start implementing some of those changes. this is heavy work for the front end

Comment (mmnoo) [reply src/components/pages/collectRecordFormPages/HabitatComplexityForm/HabitatComplexityForm.js:120]: ooooooooo

File: src/components/pages/collectRecordFormPages/HabitatComplexityForm/HabitatComplexityForm.js [hunk 2]
```diff
@@ -114,8 +120,22 @@ const HabitatComplexityForm = ({ isNewRecord }) => {
...
+
+    // { name: siteName, number: transectNumber, label }
+
+    setSubNavNode(
```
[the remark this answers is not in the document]
julia-rom: haven't tested yet, but looking at the code, it seems like we are setting the sites data in the initial `useEffect` call and not accounting for future changes. so if any sites data (such as the name) is updated `onSave`, the habitat complexity component will still be relying on the initial state. We would have to fetch new sites data in any update, and specifically check if the site name was updated before doing so. lmk if that makes senses!

Adding the site name to the event info should be a simple fix in the api. we can still send the id - however it should be explicitly stated which is the id and which is the site name, instead of the generic `site` title.  while I understand there are tradeoffs we have to make for a data syncing app, I don't think this is one of them.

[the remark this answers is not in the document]
julia-rom: agree that we can make the `getDataForSubNavNode` reusable for more components!

PR #766, opened 2023-03-17
946: Collect titles refresh bug

julia-rom:
[trello ticket](https://trello.com/c/MaBgWnf4/946-collect-record-title-sub-nav-node-title-dont-refresh-when-inputs-updated-and-saved)

previously, titles would only update in a manual refresh after saving

now, titles update automatically on save

to test:

1. add a collect record
2. add site name, transect number
3. save
4. check to see that both titles update (nav and sub nav)
Screen_Shot_2023-03-17_at_1_23_51_PM

5. update site name & transect number again with new items
6. check to see that items update again
7. repeat for each protocol

### DIFF
--- a/src/components/pages/collectRecordFormPages/BenthicLitForm/BenthicLitForm.js
+++ b/src/components/pages/collectRecordFormPages/BenthicLitForm/BenthicLitForm.js
@@ -76,7 +76,7 @@ const BenthicLitform = ({ isNewRecord }) => {
                   isNewRecord,
                   collectRecord: collectRecordResponse,
                   sites: sitesResponse,
-                  protocol: collectRecordResponse.data.protocol,
+                  protocol: collectRecordResponse?.data.protocol,
                 }),
               )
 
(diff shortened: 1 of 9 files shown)


src/components/pages/collectRecordFormPages/BenthicLitForm/BenthicLitForm.js
```diff
@@ -70,18 +71,18 @@ const BenthicLitform = ({ isNewRecord }) => {
...
-                      sitesResponse,
-                      collectRecordResponse.data.protocol,
-                    )
-                  : { name: language.protocolTitles.benthiclit }
+              setSubNavNode(
+                getDataForSubNavNode({
+                  isNewRecord,
+                  collectRecord: collectRecordResponse,
+                  sites: sitesResponse,
+                  protocol: 'benthiclit',
```
mmnoo, line 79: you can also do collectRecordResponse.data.protocol instead of hardcoding this (makes it copy paste friendly and avoids hardcoding something that should ideally be in a constant/enum)

mmnoo (approved):
Tests pass. Code looks fantastico.

PR #767, opened 2023-03-17
fix: 985 display only one decimal place for interval in the observation table to avoid weird floating point math

mmnoo:
Steps to test:
- Create a new benthic PIT sample unit.
- Set the Interval Size to 0.1
- Add some observations (can be blank, just tab through the fields)

Expected:
Interval column starts at Interval Start and increases by Interval Size value. In this case 0.1
0.1
0.2
0.3
0.4
0.5

JS wont support giant numbers, but I tested this with the earth's circumference in meters (40070000) with an added decimal  and numbers of similar magnitude (a few orders higher) and it works.

I initially coded the interval size input to not allow more than one decimal place but it opened a can of worms and also made for some strange ux when typing. So the user can type a ton of decimal places, but the observation interval will only show one decimal place (confirmed with Kim that that only one decimal place is what the front end shoudl have). 

Tradeoff is that one decimal place will always be showing now (to keep the code not expensive)
...
[ticket](https://trello.com/c/YUwyRKZD/985-rounding-error-javascript-floating-point)

### DIFF
--- a/src/App/integrationTests/collectRecords/benthicPit/App.ObservationIntervalBenthicPit.test.js
+++ b/src/App/integrationTests/collectRecords/benthicPit/App.ObservationIntervalBenthicPit.test.js
@@ -41,10 +41,10 @@ test('Benthic Pit observations: intervals are derived from interval start and in
     const observationIntervalLabelsAfterFourRowsAdded =
       within(observationsSection).getAllByLabelText('Interval')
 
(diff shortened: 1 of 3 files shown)


mmnoo:
[Rounding error (Javascript floating point ?)](https://trello.com/c/YUwyRKZD/985-rounding-error-javascript-floating-point)

PR #769, opened 2023-03-20
chore: code improvements from abandoned ticket # 970

mmnoo:
Whats going on in this PR? Well, we are rethinking the pull sync errors stuff, so I didnt push my work for that, but part of that work involved:
- addressing some previous PR stuff as promised 
- updating some var names to be better

No testing required. This is just refactoring.

### DIFF
--- a/src/App/App.js
+++ b/src/App/App.js
@@ -48,13 +48,13 @@ function App({ dexieCurrentUserInstance }) {
     [logoutMermaid],
   )
 
-  const handleSyncPushErrors = (projectsWithSyncErrors) => {
+  const handleUserDeniedSyncPush = (projectsWithSyncErrors) => {
     // projectsWithSyncErrors's type: { projectId: { name: string, apiDataTablesThatRejectedSyncing: string[] } }
     Object.entries(projectsWithSyncErrors).forEach((projectWithSyncErrorsEntry) => {
       const projectId = projectWithSyncErrorsEntry[0]
       const { name: projectName, apiDataTablesThatRejectedSyncing } = projectWithSyncErrorsEntry[1]
 
-      const toastContent = (
+      const syncErrorUserMessaging = (
         <div data-testid={`sync-error-for-project-${projectId}`}>
           <P>{language.error.getPushSyncErrorMessage(projectName)}</P>
           {language.error.pushSyncErrorMessageUnsavedData}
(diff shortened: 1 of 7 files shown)


src/App/App.js
```diff
@@ -91,19 +87,12 @@ function App({ dexieCurrentUserInstance }) {
...
-      getAccessToken &&
-      handleSyncPullErrors &&
-      handleSyncPushErrors
-    ) {
+    if (dexiePerUserDataInstance && apiBaseUrl && getAccessToken && handleUserDeniedSyncPush) {
       return new SyncApiDataIntoOfflineStorage({
         dexiePerUserDataInstance,
         apiBaseUrl,
         getAccessToken,
-        handleSyncPullErrors,
```
mmnoo, line 105: I am only ok with speculative code if the next PR is going to wrap it up and finish whatever the thing is, but now I am not sure when we are doing the pull stuff, so I am ripping that out.

src/App/mermaidData/syncApiDataIntoOfflineStorage/SyncApiDataIntoOfflineStorage.js
```diff
@@ -6,8 +6,8 @@ const SyncApiDataIntoOfflineStorage = class {
 
   _dexiePerUserDataInstance
 
-  _getTableNamesWhereSyncingWasRejectedListedPerProject =
-    function _getTableNamesWhereSyncingWasRejectedListedPerProject(pushResponse) {
+  #getTableNamesWhereSyncingWasRejectedListedPerProjectFromSyncPushError =
```
mmnoo, line 9: since pull will be different... most likely. If its not I will be reusing and renaming this, but for the time being its likely to stay different  enough from pull to be explicit about it with naming to help people out a little sans typescript.

julia-rom: i'm confused by the # sign in front - are we using this convention instead of _ ?

julia-rom (approved):

mmnoo:
@julia-rom `#varName` is JS being stupid and not explicit. Its to tell JS that it is a private class member. Cant they just annotate with the word `private`??? would be easier to read. 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_field

julia-rom:
> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_field

aahh yess thanks! haven't used those in a while!

PR #775, opened 2023-03-23
Add fix to only show error specific to project

julia-rom:
[trello card](https://trello.com/c/zFaued8v/990-i-like-the-warning-message-that-inform-me-that-i-have-read-only-role-in-a-project-but-it-appears-everywhere-d)

test instructions:

1. User B (project admin) adds User A as Collector
2. User A enters a CR
3. User B changes role of User A to Read-only before User A can save (easiest way to reproduce is for User A to go offline)
4. User A navigates to any page; this triggers a push and the response contains a collect_records object with an item with 5. status_code: 403 which triggers the toast attached to this card.
5. When use B navigates to a different project page, warning no longer appears


src/App/App.js
```diff
@@ -53,22 +55,34 @@ function App({ dexieCurrentUserInstance }) {
...
-      )
-
-      toast.error(...getToastArguments(syncErrorUserMessaging))
+      //const currentUrlPath = window.location.href
+      // the test environment wont see client side routing paths (I cosoled it and looked at the console output of the test)
+      // console.log(currentUrlPath)
+      console.log(location.pathname, projectId, location.pathname.includes(projectId))
+
+      //const isErrorSpecificToProject = location.pathname.includes(projectId)
+      // the projectId here isnt necessarily the same one as in the url since we have offline ready projects which will be pushed/synced no matter what project page you are on.
```
julia-rom, line 64: excellent point

src/App/App.js
```diff
@@ -53,22 +55,34 @@ function App({ dexieCurrentUserInstance }) {
...
+      // instead we can check if there is a project id.
+
+      const isCurrentRouteAProjectPage = !!getProjectIdFromLocation(location)
+
+      if (isCurrentRouteAProjectPage) {
```
julia-rom, line 69: will this be for any project match though? what about matching the ids

src/App/App.js
julia-rom: isn't it ok if they're pushed/synced though if you're on a different project page? as long as you're only seeing the relevant one to the project page you're currently on? @mmnoo

src/App/App.js
mmnoo: ya, we dont care what project the toast if for, just dont show it on the projects list page. actually let me read the ticket
...

mmnoo: k, the ticket technically says only on project related pages, but I think you bring up a good point. @saanobhaai if you are on a project A page, do you only want to see push errors for project A even if projectB is producing errors?

julia-rom: exactly, that was my understanding. only show project toasts to the relevant project you are on

mmnoo: I wasnt thinking when I simplified....

src/App/App.js
```diff
@@ -48,29 +50,38 @@ function App({ dexieCurrentUserInstance }) {
...
+  const handleUserDeniedSyncPush = useCallback(
+    (projectsWithSyncErrors) => {
+      // projectsWithSyncErrors's type: { projectId: { name: string, apiDataTablesThatRejectedSyncing: string[] } }
+      Object.entries(projectsWithSyncErrors).forEach((projectWithSyncErrorsEntry) => {
+        const projectId = projectWithSyncErrorsEntry[0]
+        const { name: projectName, apiDataTablesThatRejectedSyncing } =
+          projectWithSyncErrorsEntry[1]
+
+        const currentPagesProjectId = getProjectIdFromLocation(location)
+        const isErrorSpecificToProject = currentPagesProjectId === projectId
```
mmnoo, line 62: you decided not to use includes anymore?

src/App/mermaidData/syncApiDataIntoOfflineStorage/useInitializeSyncApiDataIntoOfflineStorage.js
```diff
@@ -2,19 +2,10 @@ import { useEffect, useRef } from 'react'
...
-const getProjectIdFromLocation = (location) => {
-  const { pathname } = location
-
-  const pathNameParts = pathname.split('/')
-
-  const projectId = pathNameParts[pathNameParts.indexOf('projects') + 1]
-
-  return projectId
-}
-
```
mmnoo, line 17: eek, this logic was faulty because index of projects might in the future be -1 (eg if we add a new route for `/documentation` or anything), and then because -1 + 1 = 0 projectId would end up being `/documentation` and it would cause errors everywhere with syncing probably. Bad historical Melissa.

src/library/getProjectIdFromLocation.js
```diff
@@ -0,0 +1,12 @@
+export const getProjectIdFromLocation = (location) => {
+  const { pathname } = location
+  const pathNameParts = pathname.split('/')
+  const isAProjectPage = pathNameParts[1] === 'projects' && pathNameParts.length > 2
```
mmnoo, line 4: muuuch bettter :)

julia-rom:
did an additional tiny update to the tests after merging develop - all tests passing now so going to merge in @mmnoo

PR #777, opened 2023-03-29
test: 1032 1035 tests for handling push sync status codes of 500

mmnoo:
Just tests for toasts in response to the push sync status of 500


mmnoo:
Merging to make sure this test runs and because its low risk. Please still review when you get a chance

PR #778, opened 2023-04-03
1017 - refactoring how we remove stuff with a pull response

mmnoo:
Description of work:

More details in [the ticket](https://trello.com/c/Ao12yGaR/1017-refactor-sync-pull-401-or-403-subnode-response-to-remove-items-differently) and PR comments, but this basically:
- uses a new `removes` array full of ids to delete items from our offline data store (in addition to the existing `deletes` array)
- refactors the removal of data when a user is denied syncing (nested 402) to do it based on project id instead of an array or record ids the server sends with a nested error
- when a user is denied syncing, we now make sure to remove the project from having a status of 'offline ready' (NOTE: currently if we stay on a project page, the app will reinstate that status, but we will implement a redirect to the project list in ticket 922, coming soon. 922 will also include tests for the removal of offline ready status)
- we remove a hack to keep `projects` in sync (note: different than `project_profiles`) - this is now covered by the `removes` array. 

Good candidates to test are to follow the instructions in 980 and 988, but given timelines and availability and context required for understanding, maybe leave for @saanobhaai et al.


src/App/integrationTests/app.lastRevisionNumbersForSyncing.test.js
```diff
@@ -1,130 +0,0 @@
-import '@testing-library/jest-dom/extend-expect'
```
mmnoo, line 1: just a file rename

src/App/integrationTests/app.userDeniedSyncPullNullLastRevisionNumbers.test.js
```diff
@@ -0,0 +1,228 @@
...
+  expect(lastRevisionProject1CollectRecordsBeforeSyncError[0].lastRevisionNumber).toEqual('initial')
+  expect(lastRevisionProject1ManagementsBeforeSyncError[0].lastRevisionNumber).toEqual('initial')
+  expect(lastRevisionProject1ProfilesBeforeSyncError[0].lastRevisionNumber).toEqual('initial')
+  expect(lastRevisionProject1SitesBeforeSyncError[0].lastRevisionNumber).toEqual('initial')
+
+  // after we have the page loaded,
+  // we set up the mock API to have sync errors,
+  // then we navigate the user to trigger a sync
+
+  const _mockApiWithFakeServerGeneratedLastRevisionNumbers = mockMermaidApiAllSuccessful.use(
```
mmnoo, line 89: Im gunna start naming the mocks something descriptive

src/App/mermaidData/getIsDataTypeProjectAssociated.js
```diff
@@ -1,5 +1,7 @@
-export const getIsDataTypeProjectAssociated = (dataType) =>
-  dataType === 'collect_records' ||
-  dataType === 'project_managements' ||
-  dataType === 'project_profiles' ||
-  dataType === 'project_sites'
+import { apiDataTypes } from '../../library/constants/constants'
+
```
mmnoo, line 2: swapped out raw values to use enum

src/library/constants/constants.js
```diff
@@ -5,5 +5,15 @@ export const PROJECT_CODES = {
...
+  benthicAttributes: 'benthic_attributes',
+  choices: 'choices',
+  collectRecords: 'collect_records',
+  fishFamilies: 'fish_families',
+  fishGenera: 'fish_genera',
+  fishSpecies: 'fish_species',
+  projectManagements: 'project_managements',
+  projectProfiles: 'project_profiles',
+  projectSites: 'project_sites',
+}
```
mmnoo, line 18: this probably should have been an enum all along. I didnt yak shave here to refactor everything to use this, but its here for moving forward and opportunistic refactor. 

Also it looks like I forgot to freeze it. Will do.

src/App/mermaidData/pullApiData.js
```diff
@@ -45,14 +42,6 @@ export const pullApiData = async ({
     await getAuthorizationHeaders(getAccessToken),
   )
 
-  // If projects are part of the API pull determine if the user has been
```
mmnoo, line 48: this is now taken care by deleting whatever is in the `removes` array that the api sends as part of the pull response

mmnoo:
Hmm, It looks like the robot forgot to add a comment with the preview link, but https://preview.app2.datamermaid.org/778/index.html looks to have the changes.

src/App/integrationTests/app.userDeniedSyncPullDeleteProjectRelatedRecords.test.js
```diff
@@ -0,0 +1,187 @@
...
+  expect(choicesBeforeSyncError.length).toBeGreaterThan(0)
+  expect(fishFamiliesBeforeSyncError.length).toBeGreaterThan(0)
+  expect(fishGeneraBeforeSyncError.length).toBeGreaterThan(0)
+  expect(fishSpeciesBeforeSyncError.length).toBeGreaterThan(0)
+  expect(project5CollectRecordsBeforeSyncError.length).toBeGreaterThan(0)
+  expect(project5ManagementsBeforeSyncError.length).toBeGreaterThan(0)
+  expect(project5ProfilesBeforeSyncError.length).toBeGreaterThan(0)
+  expect(project5SitesBeforeSyncError.length).toBeGreaterThan(0)
+
+  // after we have the page loaded,
```
julia-rom, line 68: nice and helpful comments!

src/App/integrationTests/app.userDeniedSyncPullNullLastRevisionNumbers.test.js
```diff
@@ -0,0 +1,228 @@
...
+      .toArray()
+
+  const lastRevisionProject1SitesAfterSyncError =
+    await dexiePerUserDataInstance.uiState_lastRevisionNumbersPulled
+      .where({ dataType: 'project_sites', projectId: '1' })
+      .toArray()
+
+  // when a user is denied a sync pull, we want to not mess with the last revision numbers for untouched 'tables'
+  expect(lastRevisionProject1BenthicAttributesAfterSyncError[0].lastRevisionNumber).toEqual(
+    'server',
```
julia-rom, line 216: why does the` lastRevisionNumber `equal 'server' ?

src/App/mermaidData/pullApiData.js
```diff
@@ -89,55 +79,40 @@ export const pullApiData = async ({
...
-          if (deleteProjectIds.length) {
-            // Delete the projects from IndexedDB as the user has been removed from them
-            dexiePerUserDataInstance.projects.bulkDelete(deleteProjectIds)
+          if (is401Or403 && isDataTypeProjectAssociated && projectId) {
+            // we still delete project related data in addition to anything in the removes array,
+            // because the removes array response doesnt include all the things we want to delete currently.
+            const deleteProjectRelatedRecords = dexiePerUserDataInstance[apiDataType]
+              .where({ project: projectId })
+              .delete()
+            const resetProjectRelatedLastRevisionNumbers =
```
julia-rom, line 101: if a project is deleted/removed, are we resetting the last revision to one prior to deletion? is there a new revision number added upon deletion?

julia-rom (approved):
tests pass, also really like that you added comments to the tests. had a couple questions for code clarification. other than that looks good

src/App/integrationTests/app.userDeniedSyncPullNullLastRevisionNumbers.test.js
mmnoo: We're testing that these types of data dont have their last revision numbers wiped. the mock api above sets the last revision numbers to server, so we want to see that in the browser, we have kept whatever the server has sent.

For the other tables below, we want to have wiped last revision number, so that next time we pull, we get _everything_, not just what ever has changed on the api. Last revision number is for the api to determine how uptodate/ in sync the browser is and it only sends changes. Usually the front end doesnt touch last revision numbers, it just stores whatever the api sends, except here where it wipes last revision numbers after we wipe the data corresponding to the last revision numbers.

src/App/mermaidData/pullApiData.js
mmnoo: this is the weird part. The `deletes` array is for data that is deleted as data. HEre, a user is just not allowed on the project. So nothing is `deleted` as data, but de do want to remove access from the user (so we do delete it from the browser). Next time that user pulls and they do have permission, they will be pulling everything from a fresh slate (because we deleted everything for them when they were denied), so we just null out their last revision number to tell the server to give us everything (everything relavent) for that project+user

mmnoo: TBH, its a bit complicated for the front end. Ideally, the API would reset last revision numbers when a user is denied and just give us waht we need to remove in the `removes` array. The front end should only concern itself with storing whatever last revision number the server sends. But this is how it evolved/currently works

src/App/integrationTests/app.userDeniedSyncPullNullLastRevisionNumbers.test.js
julia-rom: ah ok that makes sense - we just want to ensure something is there. thanks for explaining!

src/App/mermaidData/pullApiData.js
julia-rom: ahh got it

PR #784, opened 2023-04-06
Melissa/1016 handle nonresponsive server

mmnoo:
[ticket](https://trello.com/c/gJ684ORe/1016-front-end-handle-when-server-doesnt-respond-to-any-request-eg-internet-cut-out)

Description:
when a http request throws an error because the server was unresponsive, show a toast and put the app into offline mode (dont show a toast if the request is a background process and not user-initiated). I also audited the app (searching all the caught promises) to make sure we are using the http handler where appropriate. 

To test (locally). 
- Run the server, run the front end.
- go to any page while the server is running
- do something (navigating to a project-related page), and while the server is computing, kill the server

Expected results:
A toast with 'Unable to communicate with server. Changes saved on your computer, but not online.' shows
The app goes into online mode

More test:
after the above, restart the server and wait a while. The app shoudl return to online mode after it pings the server (a 'health check')

### DIFF
--- a/src/App/App.js
+++ b/src/App/App.js
@@ -46,7 +46,7 @@ function App({ dexieCurrentUserInstance }) {
     dexieCurrentUserInstance,
   })
 
-  const handleHttpResponseErrorPartiallyApplied = useCallback(
+  const handleHttpResponseErrorWithLogoutAndSetServerNotReachableApplied = useCallback(
     ({ error, callback, shouldShowServerNonResponseMessage }) =>
       handleHttpResponseError({
         error,
@@ -109,7 +109,7 @@ function App({ dexieCurrentUserInstance }) {
     isMermaidAuthenticated,
     isAppOnline,
     isSyncInProgress,
-    handleHttpResponseErrorPartiallyApplied,
+    handleHttpResponseErrorWithLogoutAndSetServerNotReachableApplied,
   })
 
   const { dexiePerUserDataInstance } = useDexiePerUserDataInstance({
@@ -143,7 +143,7 @@ function App({ dexieCurrentUserInstance }) {
     dexiePerUserDataInstance,
     isMounted,
     isAppOnline,
-    handleHttpResponseError: handleHttpResponseErrorPartiallyApplied,
+    handleHttpResponseError: handleHttpResponseErrorWithLogoutAndSetServerNotReachableApplied,
     syncApiDataIntoOfflineStorage: apiSyncInstance,
   })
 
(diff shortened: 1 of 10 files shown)


mmnoo:
[Front end handle when server doesnt respond to any request (eg internet cut out)](https://trello.com/c/gJ684ORe/1016-front-end-handle-when-server-doesnt-respond-to-any-request-eg-internet-cut-out)

src/App/App.js
```diff
@@ -46,9 +46,16 @@ function App({ dexieCurrentUserInstance }) {
     dexieCurrentUserInstance,
   })
 
-  const handleHttpResponseErrorWithLogoutFunction = useCallback(
-    ({ error, callback }) => handleHttpResponseError({ error, callback, logoutMermaid }),
-    [logoutMermaid],
+  const handleHttpResponseErrorPartiallyApplied = useCallback(
+    ({ error, callback, shouldShowServerNonResponseMessage }) =>
```
julia-rom, line 50: what makes this partially applied?

julia-rom (commented):
tested, works well. Would like to see clearing naming on `handleHttpResponseErrorPartiallyApplied` . what makes this partially applied? can we specify it in the var name?

mmnoo:
> tested, works well. Would like to see clearing naming on `handleHttpResponseErrorPartiallyApplied` . what makes this partially applied? can we specify it in the var name?

Hmm, ya, I need to think on that one. I renamed it because its not just 'withLogout' anymore, but I also want to differentiate it from the orginal function that isnt partially applied (also because auto import can cause a hassle when we are trying to use it).  Where it is used, all over the app, all those details are distracting and not required (its kind of imperative info everywhere but in app.js, and any time we change any internal workings (like adding a new param), I dont want to refactor  and have a huge diff, but I think it might be better to just be explicit with imperative info, and have to refactor (not a huge dealio).

src/App/App.js
julia-rom: ah ok ! maybe throw in a comment above the function to clarify?

>>. we're 'partially applying' the function with some but not all params : logoutMermaid and setServerNotReachable

julia-rom:
> > tested, works well. Would like to see clearing naming on `handleHttpResponseErrorPartiallyApplied` . what makes this partially applied? can we specify it in the var name?
> 
> Hmm, ya, I need to think on that one. I renamed it because its not just 'withLogout' anymore, but I also want to differentiate it from the orginal function that isnt partially applied (also because auto import can cause a hassle when we are trying to use it). Where it is used, all over the app, all those details are distracting and not required (its kind of imperative info everywhere but in app.js, and any time we change any internal workings (like adding a new param), I dont want to refactor and have a huge diff, but I think it might be better to just be explicit with imperative info, and have to refactor (not a huge dealio).

I think just adding a comment would work well here, no need to refactor (approved otherwise!)

mmnoo:
I liked your suggestion so went with that. Better to maintain an possibly over descriptive var name than a comment (which is more likely to get stale)

PR #788, opened 2023-04-17
Add disabled number field to bleaching transect inputs + Required indicator to MR rules

julia-rom:
note : wanted to make these as separate prs but accidentally made a commit to the same branch so just leaving it since they're both relatively small tasks

[trello card 1](https://trello.com/c/ocsQTJbZ/1010-add-disabled-number-field-to-bleaching-cr)
[trello card 2](https://trello.com/c/LZNZRIFR/1031-mr-rules-field-should-have-a-red-asterisk)

**Summary**
task 1: add disabled Number field to Bleaching CR - labeled Number and appearing above Label.
task 2: MR Rules field name should have a red asterisk to indicate it’s required

to test:

first task
1. go to new bleaching collect record
2. ensure number field exists and is disabled

second task
1. go to management rules
2. add new rule
3. check to see that rules has an asterisk

### DIFF
--- a/src/components/pages/collectRecordFormPages/BleachingForm/BleachingTransectInputs.js
+++ b/src/components/pages/collectRecordFormPages/BleachingForm/BleachingTransectInputs.js
@@ -14,7 +14,6 @@ import InputRadioWithLabelAndValidation from '../../../mermaidInputs/InputRadioW
 import InputWithLabelAndValidation from '../../../mermaidInputs/InputWithLabelAndValidation'
 import TextareaWithLabelAndValidation from '../../../mermaidInputs/TextareaWithLabelAndValidation'
 
-const NUMBER_VALIDATION_PATH = 'data.quadrat_collection.number'
 const CURRENT_VALIDATION_PATH = 'data.quadrat_collection.current'
 const DEPTH_VALIDATION_PATH = 'data.quadrat_collection.depth'
 const LABEL_VALIDATION_PATH = 'data.quadrat_collection.label'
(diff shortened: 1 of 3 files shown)


src/components/pages/collectRecordFormPages/BleachingForm/BleachingTransectInputs.js
```diff
@@ -149,6 +157,23 @@ const BleachingTransectInputs = ({
...
+            ignoreNonObservationFieldValidations({ validationPath: NUMBER_VALIDATION_PATH })
+          }}
+          resetNonObservationFieldValidations={() => {
+            resetNonObservationFieldValidations({ validationPath: NUMBER_VALIDATION_PATH })
+          }}
+          {...labelValidationProperties}
+          onBlur={formik.handleBlur}
+          value={formik.values.number}
+          onChange={handleNumberChange}
+        />
```
mmnoo, line 176: I think we can get rid of all the props other than:
- disabled
- id
- testId
- type
-value
-lable

src/components/pages/collectRecordFormPages/BleachingForm/BleachingTransectInputs.js
```diff
@@ -84,6 +85,13 @@ const BleachingTransectInputs = ({
     areValidationsShowing,
   )
 
+  const handleNumberChange = (event) => {
+    formik.handleChange(event)
+    resetNonObservationFieldValidations({
+      validationPath: NUMBER_VALIDATION_PATH,
+    })
+  }
+
```
mmnoo, line 94: if its always disabled, why a handler?

src/components/pages/collectRecordFormPages/BleachingForm/BleachingTransectInputs.js
julia-rom: I would prefer to leave some of these - we are using `disabled`. also` id` should be there for html context and screen readers. 

not sure how` testId` is used, and it is not a required prop so may its fine to remove this . `type `doesn't seem to be listed in the `propTypes` for  `InputWithLabelAndValidation`

mmnoo: I forgot to say, conditional approval on minimizing based on comments. :)

PR #791, opened 2023-04-22
feat: 454-related. This isnt a bug 'fix' likely, but it when a user toggles back online after being offline and the internet or server is unresponsive, the app will find out much faster and it will turn itself back offline.

mmnoo:
**Description:** There is a bug I cant reproduce (you can see some of my attempts in the ticket comments). Basically it was observed by Emily out in the real world with a crap internet connection that when returning online the user gets stuck on a loading screen. Emily tried to fix it by immediately turning the app offline, and it was still stuck. All I observed was a slow loader when turning the app back online (it waits for the health check to realize the internet is crapped out). to at least speed that up for UX reasons, not to fix the bug Emily saw:
- is to give axios a timeout for health checks, 
- and also to manually trigger a health check when the user toggles online (to they dont have to wait until the next scheduled one)

**Steps to test**
1. run app locally, point it to remote dev server (https://dev-api.datamermaid.org/v1)
2. set the app to offline mode manually
3. unscrew cable internet from router
4. turn the app online manually, it should be online for a moment before turning itself back offline. The app should load and operate like normal. 
5. you should not be able to turn the app online again until the next healthcheck determines that the internet/server is operating (after you plug your internet cable back into your modem)


src/library/onlineStatusContext.js
```diff
@@ -17,24 +17,25 @@ const OnlineStatusProvider = ({ children, value }) => {
...
-  const stopPingingApi = useCallback(() => {
-    setIsServerReachable(null)
-    clearInterval(rePingApiRef.current)
-  }, [])
+  const pingApi = () => {
+    axios
+      .get(`${apiBaseUrl}/health/`, {
+        cache: false,
+        method: 'HEAD',
+        timeout: 2000,
```
mmnoo, line 25: timeout here

src/library/onlineStatusContext.js
```diff
@@ -17,24 +17,25 @@ const OnlineStatusProvider = ({ children, value }) => {
 
   const rePingApiRef = useRef()
 
-  const stopPingingApi = useCallback(() => {
```
mmnoo, line 20: I dont know why this is in a callback.... it shoudl live inside the effect. Maybe the code evolved...

src/library/onlineStatusContext.js
mmnoo: moved this outside of the useEffect so we can manually call it when the user toggles online.

src/library/onlineStatusContext.js
```diff
@@ -50,9 +51,17 @@ const OnlineStatusProvider = ({ children, value }) => {
     return () => {
       stopPingingApi()
     }
-  }, [isNavigatorOnline, stopPingingApi])
+  }, [isNavigatorOnline])
 
   const toggleUserOnlineStatusOverride = () => {
+    const isTheUserAboutToTurnTheAppOnline = hasUserTurnedAppOffline // this is still old state, we havent updated anything yet, thats why its the opposite of what you would expect.
```
mmnoo, line 57: this is counterintuitive because the `hasUserTurnedAppOffline` is about to be/is stale state so it reads funny

mmnoo:
[stuck returning to online but no network](https://trello.com/c/LjTaPOwZ/454-stuck-returning-to-online-but-no-network)

saanobhaai:
OK so I agree on not being able to reproduce; I tried everything I could think of to get what Emily got, using dev-app, and couldn't. Because of your 1016 work, as soon as the health check fails, it goes into offline mode (and this is awesome). I also can't quite make out the version number on her video, but I think what happened is that she was using v2.1.0, went offline, I pushed v2.2.0 with 1016 merged, and then she had her incident. Just a few days too late, in other words.

One thing I did in testing was to try different throttling settings, thinking maybe she was able to click the offline toggle before the health check had completed. I'm fairly confident that in addition to her hotspot issues, the connection was slow or had high latency (or both). It's *possible* the scenario included a health check that was successful (because small/quick) before the connection dropped, allowing her to manually click offline, but I doubt it -- and even if so, the 1016 work, if present, would still have meant converting to offline and allowing her to continue to work.

So, I think we should:
- consider the root problem fixed by 1016, as you suggest. I'll circle back with Emily on this.
- include this work as a belt-and-suspenders improvement -- I like doing the health check right away.
- NOT add the health check timeout. When in Chrome's devtool's "Slow 3G" throttling profile, the health check always failed, so I could never get online. ("Fast 3G" seemed to be ok, I think because of lower latency.)

src/library/onlineStatusContext.js
mmnoo: Julia, I removed the timeout based on Kim's comment about slow 3g and it giving false fails.

saanobhaai (approved):

src/library/onlineStatusContext.js
julia-rom: how do we know if the user is about to turn the app online? are we referring to the app connection state here? if so can we make the var more specific to that? `isPoorHealthCheckDetected`

julia-rom (approved):
just some clarity on var names

PR #795, opened 2023-05-03
Remove disabled fish belt bin sizes - add clear sizes modal

julia-rom:
[trello card](https://trello.com/b/XeUt1nb5/mermaid-dev)

Users should be able to change the fish size bin even if there are observations. Show a modal confirmation that the "Size" value will be cleared for all observations.

To test:
1. add a new fish belt record
2. select a bin size
3. add observations
4. change bin size
5. modal should appear with a warning that sizes will be cleared if they accept
6. check to see that sizes are cleared if you click ok

### DIFF
--- a/src/components/pages/collectRecordFormPages/FishBeltForm/FishBeltTransectInputs.js
+++ b/src/components/pages/collectRecordFormPages/FishBeltForm/FishBeltTransectInputs.js
@@ -75,6 +75,7 @@ const FishBeltTransectInputs = ({
   const hasFishBeltObservations =
     !!observationsState.length > 0 && observationsState[0]?.fish_attribute
   const [isClearSizeValueModalOpen, setIsClearSizeValueModalOpen] = useState(false)
+  const [sizeBinEvent, setSizeBinEvent] = useState({})
 
   const transectNumberValidationProperties = getValidationPropertiesForInput(
     fishbelt_transect?.number,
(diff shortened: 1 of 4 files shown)


src/components/pages/collectRecordFormPages/FishBeltForm/FishBeltTransectInputs.js
```diff
@@ -72,6 +74,8 @@ const FishBeltTransectInputs = ({
   // account for empty starter row
   const hasFishBeltObservations =
     !!observationsState.length > 0 && observationsState[0]?.fish_attribute
+  const [isClearSizeValueModalOpen, setIsClearSizeValueModalOpen] = useState(false)
+  const [sizeBinEvent, setSizeBinEvent] = useState({})
```
mmnoo, line 78: Its weird that we are storing an whole event, but I get you on refactor tradeoffs. (Im actually passing around whole events on my other project for timeline tradeoffs, but I dont like it)

mmnoo (approved):
Tested. Looks great! I like the UX balance of tradeoffs made.

src/components/pages/collectRecordFormPages/FishBeltForm/FishBeltTransectInputs.js
julia-rom: yeah it feels like a tradeoff but since we're using it in `resetSizeValues` function, decided it was worth keeping for now

PR #806, opened 2023-06-05
cursor fix on data sharing

julia-rom:
- `input` and `label `were overriding the cursor style set by `MaxWidthInputWrapper `
-  had to set these individually to ensure consistent cursor styling as mentioned in this [trello card](https://trello.com/c/FKkSLucU/382-incorrect-cursor-on-data-sharing-page)

to test:
1. go to data sharing
2. ensure pointer cursor is above radio buttons
3. when updating, it switches to wait
4. same for checkbox label

### DIFF
--- a/src/components/pages/DataSharing/DataSharing.js
+++ b/src/components/pages/DataSharing/DataSharing.js
@@ -49,18 +49,18 @@ const DataSharingTable = styled(Table)`
 `
 const CheckBoxLabel = styled.label`
   display: inline-block;
-  cursor: ${(props) => props.cursor || 'pointer'};
+  cursor: ${(props) => props.cursor};
   input {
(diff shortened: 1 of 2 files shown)


src/components/pages/DataSharing/DataSharing.js
```diff
@@ -49,11 +49,20 @@ const DataSharingTable = styled(Table)`
 `
 const CheckBoxLabel = styled.label`
   display: inline-block;
+  cursor: ${(props) => props.cursor};
   input {
     margin: 0 ${theme.spacing.xsmall} 0 0;
   }
 `
 
+const Input = styled.input`
```
julia-rom, line 58: both input and label causing cursor overrides - set these individually

PR #809, opened 2023-06-09
[M396] DRYed out biomass calc styles

alanjleonard:


alanjleonard:
[Make it clear that the biomass calculation in the observations table isn't editable.](https://trello.com/c/Tr92tfHJ/396-make-it-clear-that-the-biomass-calculation-in-the-observations-table-isnt-editable)

julia-rom (approved):

PR #812, opened 2023-06-14
[360] Cursor fix for table on users page

julia-rom:
[trello ticket](https://trello.com/c/XwCyweuS/360-mouse-shows-clickable-anywhere-on-submitted-row-but-clicking-only-goes-to-su-in-the-method-column)

note: other tables were fixed in a separate pr - but good to cross check just incase:

collecting
submitted
Sites
MRs
Users
Observations

to test:
1. go to users page table
2. hover over radio buttons
3. pointer cursor shows over labels/inputs
4. auto cursor shows when button is disabled 
5. wait cursor shows when updating

### DIFF
--- a/src/components/pages/Users/Users.js
+++ b/src/components/pages/Users/Users.js
@@ -111,7 +111,7 @@ const TableRadioLabel = styled.label(
   (props) => css`
     top: 0;
     right: 0;
-    cursor: ${props.disabled ? 'auto' : 'pointer'};
+    cursor: ${props.cursor};
     bottom: 0;
     left: 0;
     position: absolute;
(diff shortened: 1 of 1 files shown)


src/components/pages/Users/Users.js
```diff
@@ -107,18 +107,27 @@ const NameCellStyle = styled('div')`
...
+    position: absolute;
+    display: grid;
+    place-items: center;
+    ${hoverState(css`
+      border: solid 1px ${theme.color.primaryColor};
+    `)}
+  `,
+)
+
+const TableRadioInput = styled.input(
```
julia-rom, line 126: both `label` and `input` need a cursor selection - setting it on `label `does not overwrite it on `input`

tienhoah: could `input` stay inside `TableRadioLabel` now. It looks much cleaner without passing `getCursorType()` twice into 2 styled components? What do you think? This is up to you.

julia-rom: added! good idea

PR #813, opened 2023-06-14
[496] disable observations if bin size not selected

julia-rom:
[trello ticket ](https://trello.com/c/aqNCQdit/496-still-able-to-add-observations-to-a-fish-belt-if-fish-size-bin-isnt-selected)

to test:

1. add a fish belt collect record
2. try and add an observation without selecting a bin size
3. add-row should be disabled, in addition to any text inputs
4. tab and enter should also be disabled to add a new row


tienhoah (approved):
Looks good here, even though the preview link is not up to date (I test merging my pr for switching radio to select dropdown).

PR #814, opened 2023-06-15
[495] fix save prompt for observation inputs on collect record pages

julia-rom:
[trello card](https://trello.com/c/b68S3yEH/495-collect-records-dont-prompt-to-save-when-navigating-away-when-theres-only-observations)

to test:
1. go to a collect record page
2. add an observation input
3. navigate away from page
4. ensure save prompt shows up

### DIFF
--- a/src/components/pages/collectRecordFormPages/CollectRecordFormPage/CollectRecordFormPage.js
+++ b/src/components/pages/collectRecordFormPages/CollectRecordFormPage/CollectRecordFormPage.js
@@ -616,7 +616,7 @@ const CollectRecordFormPage = ({
       />
 
       {displayLoadingModal && <LoadingModal />}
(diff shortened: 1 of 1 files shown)


src/components/pages/collectRecordFormPages/CollectRecordFormPage/CollectRecordFormPage.js
```diff
@@ -616,7 +616,7 @@ const CollectRecordFormPage = ({
       />
 
       {displayLoadingModal && <LoadingModal />}
-      <EnhancedPrompt shouldPromptTrigger={formik.dirty} />
+      <EnhancedPrompt shouldPromptTrigger={formik.dirty || areObservationsInputsDirty} />
```
julia-rom, line 619: observation inputs are not included in formik

PR #815, opened 2023-06-26
376: Add methods filter collect page

julia-rom:
[trello ticket](https://trello.com/c/wOlKJC46/376-add-the-method-filter-on-tables-from-v1)

Methods filter added to collect records page - multiselect drop down which works alongside global search

notes:
- using MUI component library for multiselect checkbox
- need to drill down further to get hoverstate matching theme color (will add a separate ticket)
- additional separate ticket added to update pagination counter so when filters are added (global and method drop down), table record count will be updated as well in the pagination counter
- added persist method filter

to test:
- go to collect records page table
- select items to filter in methods checkbox
- refresh, ensure filters are still checked and table is updated

intended behviour:
- all selected, show all records
- if none are selected, show all records (this is the default, which implies filter is inactive)

EDIT - working on fixing tests

### DIFF
--- a/src/components/generic/Table/usePersistUserTablePreferences.js
+++ b/src/components/generic/Table/usePersistUserTablePreferences.js
@@ -4,29 +4,35 @@ import usePrevious from '../../../library/usePrevious'
 
 const usePersistUserTablePreferences = ({ key, defaultValue = undefined }) => {
   // Persist the table user preferences in sessionStorage
-  const [tableUserPreferences, setTableUserPreferences] = useSessionStorage({ key, initialValue: defaultValue })
+  const [tableUserPreferences, setTableUserPreferences] = useSessionStorage({
+    key,
+    initialValue: defaultValue,
+  })
   // Set the previous value
   const previousPreferences = usePrevious(tableUserPreferences)
 
-  const handleSetTableUserPreferences = useCallback(({ propertyKey, currentValue }) => {
-    const previousPreferencesPropExists = previousPreferences
-      && Object.prototype.hasOwnProperty.call(previousPreferences, propertyKey)
-    const valueChanged = !previousPreferencesPropExists
-      || (
+  const handleSetTableUserPreferences = useCallback(
+    ({ propertyKey, currentValue }) => {
+      const previousPreferencesPropExists =
+        previousPreferences &&
+        Object.prototype.hasOwnProperty.call(previousPreferences, propertyKey)
+      const valueChanged =
+        !previousPreferencesPropExists ||
         // Current value sometimes set to undefined by useTable after setting it.
         // Do not update preferences when this happens
-        currentValue !== undefined
-        && JSON.stringify(currentValue) !== JSON.stringify(previousPreferences[propertyKey])
-      )
+        (currentValue !== undefined &&
+          JSON.stringify(currentValue) !== JSON.stringify(previousPreferences[propertyKey]))
 
-    if (valueChanged) {
-      // Update the preference specified in the key
-      setTableUserPreferences({
-        ...tableUserPreferences,
-        [propertyKey]: currentValue || ''
-      })
-    }
-  }, [tableUserPreferences, setTableUserPreferences, previousPreferences])
+      if (valueChanged) {
+        // Update the preference specified in the key
+        setTableUserPreferences({
+          ...tableUserPreferences,
+          [propertyKey]: currentValue || '',
+        })
+      }
+    },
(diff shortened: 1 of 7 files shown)


src/components/generic/Table/usePersistUserTablePreferences.js
```diff
@@ -4,29 +4,35 @@ import usePrevious from '../../../library/usePrevious'
...
 
-    if (valueChanged) {
-      // Update the preference specified in the key
-      setTableUserPreferences({
-        ...tableUserPreferences,
-        [propertyKey]: currentValue || ''
-      })
-    }
-  }, [tableUserPreferences, setTableUserPreferences, previousPreferences])
+      if (valueChanged) {
```
julia-rom, line 26: I didn't actually change anything in this file, vscode might have updating the spacing when I was looking at it

src/components/pages/Collect/Collect.js
```diff
@@ -55,6 +56,9 @@ const Collect = () => {
   const isMounted = useIsMounted()
   const handleHttpResponseError = useHttpResponseErrorHandler()
   const isReadOnlyUser = getIsUserReadOnlyForProject(currentUser, projectId)
+  const [methodsFilteredTableCellData, setMethodsFilteredTableCellData] = useState([])
+  const tableSessionStorage = JSON.parse(window.sessionStorage[`${currentUser.id}-collectTable`])
+  const [methodsFilter, setMethodsFilter] = useState(tableSessionStorage?.methodsFilter)
```
julia-rom, line 61: If I comment out line 61, tests will pass, however, I need this to ensure persist works on the methodsFilter (which is why I am setting the state from` tableSessionStorage`)

I decided to implement it outside of reactTable, which does have a `columnFilters` prop, but it's mostly for a string instead of an array. The documentation does not make it clear how to implement this with different cases

src/components/pages/Collect/Collect.js
```diff
@@ -136,33 +140,46 @@ const Collect = () => {
...
-          ),
-          depth: uiLabels.depth,
-          sampleDate: uiLabels.sampleDate,
-          observers: uiLabels.observers,
-        }
-      }),
-    [collectRecordsForUiDisplay, currentProjectPath],
-  )
+  const tableCellData = useMemo(() => {
+    const preparedTableCellData = collectRecordsForUiDisplay.map(({ id, data, uiLabels }) => {
```
julia-rom, line 144: made this into a var returned at the end of the code block

src/components/pages/Collect/Collect.js
```diff
@@ -136,33 +140,46 @@ const Collect = () => {
...
+            {uiLabels.size} {isQuadratSampleUnit && uiLabels.size !== noLabelSymbol && <sup>2</sup>}
+          </>
+        ),
+        depth: uiLabels.depth,
+        sampleDate: uiLabels.sampleDate,
+        observers: uiLabels.observers,
+      }
+    })
+
+    setMethodsFilteredTableCellData(preparedTableCellData)
```
julia-rom, line 167: this was so I could add line 167

src/components/pages/Collect/MethodsFilterDropDown.js
```diff
@@ -0,0 +1,92 @@
+/* eslint-disable import/no-extraneous-dependencies */
+import React from 'react'
+import PropTypes, { string } from 'prop-types'
+import OutlinedInput from '@mui/material/OutlinedInput'
```
julia-rom, line 4: not sure why these warnings crop up when `@mui/material` is indeed listed in the `package json` dependencies. Maybe it's getting confused cause we have two?

src/components/pages/Collect/Collect.js
mmnoo: I updated it to use the existing table preferences persisted state (see most recent commit) but that might have not been the crux of it. When I loaded it up on my machine, it complained about not being able to parse undefined ( ```const tableSessionStorage = JSON.parse(window.sessionStorage[`${currentUser.id}-collectTable`])```) Tests errors also hinted at the same but less explicitly. When I fixed that and the-not-able-to-read-from-undefined errors that cascaded out from my fix, there were still issues that I didnt pursue or assess. So I am wondring if your dev environment/browser always had state that it needed, but the tests complained because they always (for this scenario at least) started from a blank state.

julia-rom: I originally had it in the table, and initiated the` methods `state just before that, however, I was still seeing a bunch of failing tests. That's interesting that you're getting that undefined parsing error, as I didn't get that before!

PR #822, opened 2023-07-05
implement Methods filter on MRO table [debugging]

julia-rom:
Added methods filter on management regimes overview table similarly to Collect, Submitted, Users and Transects table

current bug:
1. go to MRO table
2. add a filter
3. refresh
4. get `TypeError:  Cannot read properties of undefined (reading 'props')`

my suspicion is that there may be another race condition happening here, but i'm not 100% sure. The type error is happening because` tableCellData` is missing data (only `site` and `method` appear, without `sampleUnitNumbers`, ie those long ids in the object you see below)

normal `tableCellData`

Screen Shot 2023-07-05 at 5 36 13 PM

what `tableCellData` looks like when its logged in the `_applyMethodsFilterOnLoad` function

Screen Shot 2023-07-05 at 5 35 46 PM

my guess is that line 187  `...populateSampleUnitNumbersByManagementRegimeRow(record)` isn't finished populating before it is used within `_applyMethodsFilterOnLoad`. Not sure why, however, as this is similar to `UsersAndTransects`

### DIFF
--- a/src/components/pages/ManagementRegimesOverview/ManagementRegimesOverview.js
+++ b/src/components/pages/ManagementRegimesOverview/ManagementRegimesOverview.js
@@ -355,8 +355,8 @@ const ManagementRegimesOverview = () => {
                 (cell) => cell.value !== '-',
               )
 
-              const mrTransectNumberRowCellValues = mrTransectNumberRowCellsWithNonEmptyValue.map(
-                (cell) => cell.value.props.sampleUnitNumbersRow.length,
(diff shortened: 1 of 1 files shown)


mmnoo:
k, those error messages were misleading. I did a hocus pocus approach and put `?.` in a bunch of places where it wasnt complaining. 

Im not sure that is the _best_ approach, but it might be the bandaid one. Maybe a better approach is to keep the loader active until the data is ready, or to just put the components/jsx that is complaining behind a ternary and return null if the data isnt present yet?

mmnoo:
Feel free to revert my commit :)

PR #823, opened 2023-07-07
[507] Bug fix for observations persist issue 

julia-rom:
[trello card](https://trello.com/c/ZT7EohIk/507-observations-disappear-when-goes-online-after-inputting-offline)

previously if you were on a new form and you entered data into observations and refreshed the page, it would get wiped out

also fixed second bug which wiped out persisted data when an observer was clicked and the page was refreshed or toggled from offline to online

to test:

// for refreshing
1. create a new collect record
2. add a few observations
3. refresh
4. observations should persist

// for offline to online
1. create a new collect record while offline
2. add observations + observer + transect inputs
3. toggle online
4. all items should persist

### DIFF
--- a/src/components/pages/collectRecordFormPages/CollectRecordFormPage/CollectRecordFormPage.js
+++ b/src/components/pages/collectRecordFormPages/CollectRecordFormPage/CollectRecordFormPage.js
@@ -65,10 +65,11 @@ function loadObservationsFromCollectRecordIntoTableState({
     })
   }
 
+  const persistedUnsavedObservationsTable = getPersistedUnsavedObservationsTableData()
+
   if (collectRecordBeingEdited && !isNewRecord) {
     const observationsFromApiTable = collectRecordBeingEdited.data[observationsPropertyName] ?? []
 
-    const persistedUnsavedObservationsTable = getPersistedUnsavedObservationsTableData()
     const initialObservationsToLoadTable =
       persistedUnsavedObservationsTable ?? observationsFromApiTable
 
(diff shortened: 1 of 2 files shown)


src/components/pages/collectRecordFormPages/CollectRecordFormPage/CollectRecordFormPage.js
```diff
@@ -87,6 +88,14 @@ function loadObservationsFromCollectRecordIntoTableState({
 
   if (isNewRecord) {
     handleAddEmptyInitialObservation()
+
+    if (persistedUnsavedObservationsTable?.length) {
```
julia-rom, line 92: we also want to account for the fact that there may be persisted observation data for new records that were unsaved

src/library/useUnsavedDirtyFormDataUtilities.js
```diff
@@ -27,14 +27,6 @@ export const useUnsavedDirtyFormDataUtilities = (sessionStorageName) => {
     }
   }, [sessionStorageName])
 
-  const _clearPersistedUnsavedFormDataBeforeUnload = useEffect(() => {
```
julia-rom, line 30: this function was breaking things so I removed it. unless someone has a good case for why it's necessary, i'm keeping it out

mmnoo: oh, quick comment before a meeting, but I think this is so that if a user navigates away, we want to clear out the persist data, otherwise the dirty form will load with stale data when the user navigates back. We only want to persist as a hack for network or whatever other react refreshes are caused by prop changes that may clear the dirty form

tienhoah: yeah, that is totally going opposite of what you are doing.

tienhoah:
I also check when changing page without saving the collect form, the session storage is cleared. (I think the part in history.listen() already handling that). 
Therefore, I think it's good here. 
Likewise, I'm not sure what is the use case for it as well.

tienhoah (approved):
Approved

mmnoo (approved):

PR #844, opened 2023-08-17
[527]  column header tooltips

julia-rom:


src/components/pages/collectRecordFormPages/HabitatComplexityForm/HabitatComplexityObservationTable.js
```diff
@@ -64,6 +68,25 @@ const HabitatComplexityObservationsTable = ({
     [intervalSize, observationsDispatch],
   )
 
+  const _useOnClickOutsideOfInfoIcon = useEffect(() => {
+    document.body.addEventListener('click', () => {
+      if (isHelperTextShowing === true) {
+        setIsHelperTextShowing(false)
+      }
+    })
+  }, [isHelperTextShowing])
```
mmnoo, line 77: mabey with adding listeners, could be to add a clean up function that clears the listener (so we dont get multiple redundant listeners - which might cause performance issues or race conditions hypothetically)

Ive been working with effects adding listeners a lot lately, and am thinking I might just clean them up instead of waiting for the inevitable bugs (in my project)not cleaning them up brings 😆

mmnoo: (otherwise approved. Not clicking approve so you are more likely to see this comment)

julia-rom: good call on removing an event listener - thinking of a good way to implement this . We could put it in a `useEffect`, but on what condition

mmnoo: I was thinking of just using that useEffects cleanup function. https://blog.logrocket.com/understanding-react-useeffect-cleanup-function/

julia-rom: interesting ... will that cancel the event listeners too? I feel like I would still need to remove the event listener in the `useEffect `return @mmnoo

PR #866, opened 2023-10-17
removed the delete icon

alanjleonard:


alanjleonard:
[Delete project main card](https://trello.com/c/P4wrqxDo/367-delete-project-main-card)

PR #869, opened 2023-10-19
M577 Bleaching collect record form text improvements

mmnoo:
Requirements from [ticket](https://trello.com/c/u6dG5qte/577-suggestion-for-bleaching-su-form):
> Change the subheading Quadrat to Quadrat Collection
Change the help text for the Number field to Number is automatically generated by MERMAID as an identifier for a bleaching sample unit, and is not editable.

I put some other text I bumped into into the language file while I was at it.


mmnoo:
[Suggestion for Bleaching SU Form](https://trello.com/c/u6dG5qte/577-suggestion-for-bleaching-su-form)

alanjleonard (approved):

src/language.js
```diff
@@ -284,6 +284,12 @@ const pages = {
     totalAbundanceLabel: 'Total Abundance',
     totalBiomassLabel: 'Total Biomass (kg/ha)',
     newBenthicAttributeLink: 'Propose New Benthic Attribute...',
+    formSectionTitle: {
+      sampleEvent: 'Sample Event',
+      quadrat: 'Quadrat Collection',
```
saanobhaai, line 289: Can we make the key `quadratcollection`? A small thing, but the distinction is what led to the current user confusion in the first place, so best to be consistent.

saanobhaai (changes requested):
One small change request

src/language.js
mmnoo: changed.

PR #870, opened 2023-10-19
M292 localhost and the dev-app subdomain should link to the dev version of dashboard. All other link to the non-dev version

mmnoo:
Basically: for localhost or `dev-app` subdomains route dashboard links to dev-dashboard.datamermaid.org, otherwise route to dashboard.datamermaid.org


.env.sample
```diff
@@ -5,5 +5,6 @@ REACT_APP_AUTH0_DOMAIN=
 REACT_APP_MERMAID_API=
 REACT_APP_MERMAID_REFERENCE_LINK=
 REACT_APP_MERMAID_DASHBOARD_LINK=
+REACT_APP_MERMAID_DEV_DASHBOARD_LINK=
```
mmnoo, line 8: Ive added this to the Dashlane secure note too.

mmnoo:
[link from app to project-specific view on dashboard](https://trello.com/c/F1x0fVuM/292-link-from-app-to-project-specific-view-on-dashboard)

saanobhaai:
This does the job and so we should probably go with it, but strikes me as odd: on the API side anyway, env vars are variable, i.e. you'd set `REACT_APP_MERMAID_DASHBOARD_LINK` in each env, and the app would use it; you wouldn't create a separate var for every env.

alanjleonard:
I think this is a good idea. I have no input on the .env file.

mmnoo:
Lol, I feel stupid for missing that obvious approach. The solution is just to twiddle the cdk stuff.  Let me do that instead of adding this extra stuff to the codebase.

PR #873, opened 2023-11-02
M272 Control if the app captures Google Analytics data with environment variables.

mmnoo:
Opt into tracking with an environment variable. `REACT_APP_CAPTURE_GOOGLE_ANALYTICS=true`. If this environment variable is not present and set explicitly to true, no tracking will occur. (our secure note for env vars has been updated)

Other options considered:
- Hostname filters within Google Analytics interface. (requires learning the nuances of that process with Google Analytics, which less of us will be able to do. Most of us can reason easily about environment vars).
- front end code to check host name and only implement analytics for `app.datamermaid.org` (which is less explicit, and tightly couples GA tracking logic to a particular host name, which feels more fragile)


saanobhaai:
[add analytics to frontend](https://trello.com/c/DYBtZzS9/272-add-analytics-to-frontend)

PR #876, opened 2023-11-09
M369 add fish name info in popover for fish belt collect record observations

mmnoo:


mmnoo:
[create popover for Reference icon in fish fields](https://trello.com/c/ra8lyEmF/369-create-popover-for-reference-icon-in-fish-fields)

PR #878, opened 2023-11-14
M500 add fish_groupings info to fish name options, and biomass calculations

mmnoo:
NOTE: this change adds a table to Dexie/IndexedDb. This means working on branches with a previous version of Dexie, after working on this one, will cause an error. Workaround: Wipe your browser storage for the app when switching to a branch with the old version. Switching to a branch with the most recent version should work without error. 

https://trello.com/c/GAogWnBm/500-add-fish-groupings


mmnoo:
[add fish groupings](https://trello.com/c/GAogWnBm/500-add-fish-groupings)

PR #880, opened 2023-11-15
M578 show explainer text for non admin users on why they can neither delete an MR or site

mmnoo:


mmnoo:
[Delete button disappeared on the site and MR page](https://trello.com/c/tUSUaxoG/578-delete-button-disappeared-on-the-site-and-mr-page)

PR #885, opened 2023-11-30
[M593] MR rule language tweak

saanobhaai:


saanobhaai:
[Add helper text to partial restrictions options.](https://trello.com/c/mfww22KP/593-add-helper-text-to-partial-restrictions-options)

mmnoo (approved):

PR #887, opened 2023-12-05
[599] Update multiselect-chip styles

julia-rom:
[trello card link](https://trello.com/c/w4faVTCM/599-styles-of-the-new-observers-multi-select-dont-match-the-rest-of-the-application)

updated drop down styles to match the method filter drop down

### DIFF
--- a/src/components/mermaidInputs/InputMuiChipSelectWithLabelAndValidation/InputMuiChipSelectWithLabelAndValidation.js
+++ b/src/components/mermaidInputs/InputMuiChipSelectWithLabelAndValidation/InputMuiChipSelectWithLabelAndValidation.js
@@ -6,7 +6,6 @@ import {
   CustomMenuItem,
   CustomMuiChip,
   CustomMuiSelect,
-  getMenuItemStyle,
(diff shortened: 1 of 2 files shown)


src/components/mermaidInputs/InputMuiChipSelectWithLabelAndValidation/InputMuiChipSelectWithLabelAndValidation.js
```diff
@@ -63,7 +62,6 @@ const InputMuiChipSelectWithLabelAndValidation = ({
               key={option.value}
               value={option.value}
               sx={{ fontSize: theme.typography.defaultFontSize }}
-              style={getMenuItemStyle({ option, selectedItems: value, theme })}
```
julia-rom, line 66: removed this because we don't use bolding in the method filter drop down - still need Al to look over it to confirm

alanjleonard:
Looks good, just two things @julia-rom 
1. Font is incorrect - color, typeface
2. Needs the `:focus` style that the other inputs have. The blue outline

julia-rom:
update chip font and added focus color

mmnoo (approved):

PR #889, opened 2023-12-06
M622 bleaching observation table jumping bug fix

julia-rom:
[trello card](https://trello.com/c/2dmq0XSL/622-bleaching-observation-table-automatically-move-up-when-typing-ba)

removed `isLastRow` for Colonies Bleached table. Left the check on Percentage Cover, since it is the last table on the page (in which case, we don't mind that it scrolls to bottom).


mmnoo (approved):
works with no bugs, and didnt break tab key behaviour

PR #891, opened 2023-12-07
M287 remove excessive network calls

mmnoo:
The API is sensitive to the absence of a trailing slash and if it is missing will cause a redirect, which results in multiple calls to  `projecttags/` and
`projecttags`
for example.

### DIFF
--- a/src/App/useInitializeCurrentUser.js
+++ b/src/App/useInitializeCurrentUser.js
@@ -10,7 +10,6 @@ export const useInitializeCurrentUser = ({
   dexieCurrentUserInstance,
   isMermaidAuthenticated,
   isAppOnline,
-  isSyncInProgress,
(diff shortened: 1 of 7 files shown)


src/App/useInitializeCurrentUser.js
```diff
@@ -51,7 +49,6 @@ export const useInitializeCurrentUser = ({
     dexieCurrentUserInstance,
     isMermaidAuthenticated,
     isAppOnline,
-    isSyncInProgress,
     handleHttpResponseErrorWithLogoutAndSetServerNotReachableApplied,
   ])
```
mmnoo, line 54: The current user was being initialized twice because this var in the dep array, which wasnt even used!!! I was hoping that might speed up tests (cause less rerendering), but no difference

julia-rom (approved):
looks good

PR #900, opened 2023-12-19
Melissa/m646 fix sites mr sort

mmnoo:
Steps to test:

GO to each of the sites and MRs page
click Export X
Expect that the resultant csv has items sorted by name


src/language.js
```diff
@@ -207,7 +207,7 @@ const createNewOptionModal = {
     <>
       Genus not found in MERMAID
       <br />
-      <a target="_blank" href="https://datamermaid.org/contact-us">
+      <a target="_blank" href="https://datamermaid.org/contact-us" rel="noreferrer">
```
mmnoo, line 210: @alanjleonard , this answers your question about noreferrer you had a while back. Linter made me do it. https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-no-target-blank.md

mmnoo:
[Sort by name for exported sites and MRs](https://trello.com/c/RoVMIBuU/646-sort-by-name-for-exported-sites-and-mrs)

src/language.js
alanjleonard: But we likely want to track where traffic is coming from. We own both domain and that link isn't dynamic. I don't see a reason for enforcing the `rel="noreferrer"`

alanjleonard: This is also a client-side security fix. If somebody was being malicious, they could just open devtools and remove that attribute.

alanjleonard: It talks about this in the `when to override it` section. https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-no-target-blank.md#when-to-override-it

src/components/pages/Sites/Sites.js
```diff
@@ -278,7 +278,7 @@ const Sites = () => {
           Notes: site.notes,
         }
       })
-      .toSorted((firstSite, secondSite) => firstSite.Name - secondSite.Name)
+      .toSorted((a, b) => a.Name.localeCompare(b.Name))
```
julia-rom, line 281: looks good

src/components/pages/ManagementRegimes/ManagementRegimes.js
```diff
@@ -239,44 +239,46 @@ const ManagementRegimes = () => {
...
+          'Secondary name': managementRegime.name_secondary,
+          'Year established': managementRegime.est_year,
+          Size: managementRegime.size,
+          Governance: governance,
+          'Estimate compliance': complianceName,
+          Rules: rules,
+          Notes: managementRegime.notes,
+        }
+      })
+      .toSorted((a, b) => a.Name.localeCompare(b.Name))
```
julia-rom, line 281: looks like line 281 was the only thing updated right? maybe an indentation triggered the block updating

julia-rom (approved):
- names are sorted, code looks good

src/language.js
mmnoo: Ya, good point. Its pointing to our own content, and we definitely want to know how many people click on the cntact us page (vs who actually contact us. Could be an interesting metric if a contact us form had a bug or bad UX in theory)

PR #903, opened 2023-12-22
Upgrade to node v20

julia-rom:
Previously, we were getting a `Error: error:0308010C:digital envelope routines::unsupported` error in the build.

This error is mentioned in the [release notes](https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V17.md#17.0.0) for Node.js 17.0.0, with a suggested workaround:

If you hit an ERR_OSSL_EVP_UNSUPPORTED error in your application with Node.js 17, it’s likely that your application or a module you’re using is attempting to use an algorithm or key size which is no longer allowed by default with OpenSSL 3.0. A command-line option, --openssl-legacy-provider, has been added to revert to the legacy provider as a temporary workaround for these tightened restrictions.


mmnoo (approved):
Well Ill be damned. We still need to ditch react-scripts, but this works for now and it better than having different node versions for build and test.

PR #909, opened 2024-01-22
[642] add sync interval start button to Benthic pit

julia-rom:
[trello card](https://trello.com/c/z70S6wyM/642-help-text-for-interval-start-in-pit-is-incomplete)

to test:

1. add new collect record
2. ensure interval start has no default value
3. add interval size number input
4. check "Use Interval Size as Interval Start"
5. notice interval start sync to interval size value and change to a disabled field
6. if unchecked, you can update both inputs individually

### DIFF
--- a/src/components/mermaidInputs/InputWithLabelAndValidation/InputWithLabelAndValidation.js
+++ b/src/components/mermaidInputs/InputWithLabelAndValidation/InputWithLabelAndValidation.js
@@ -46,7 +46,7 @@ const InputWithLabelAndValidation = ({
   useStopInputScrollingIncrementNumber(textFieldRef)
 
   const [isHelperTextShowing, setIsHelperTextShowing] = useState(false)
-  const [isChecked, setIsChecked] = useState(false)
+  const [isCheckboxChecked, setIsCheckboxChecked] = useState(false)
 
   const handleInfoIconClick = (event) => {
     isHelperTextShowing ? setIsHelperTextShowing(false) : setIsHelperTextShowing(true)
@@ -60,7 +60,7 @@ const InputWithLabelAndValidation = ({
       aria-describedby={`aria-descp${id}`}
       id={id}
       unit={unit}
-      disabled={isChecked}
+      disabled={isCheckboxChecked}
       {...restOfProps}
     />
   ) : (
@@ -73,8 +73,9 @@ const InputWithLabelAndValidation = ({
     />
   )
 
+  // example use case is in Benthic Pit Transect Inputs for Interval Start
   const handleCheckboxChange = (checked) => {
-    setIsChecked(checked)
+    setIsCheckboxChecked(checked)
     handleCheckboxUpdate(checked)
   }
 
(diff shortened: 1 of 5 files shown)


src/components/mermaidInputs/InputWithLabelAndValidation/InputWithLabelAndValidation.js
```diff
@@ -53,6 +73,11 @@ const InputWithLabelAndValidation = ({
     />
   )
 
+  const handleCheckboxChange = (checked) => {
+    setIsChecked(checked)
+    handleCheckboxUpdate(checked)
```
julia-rom, line 78: func passed in from benthic pit inputs component

src/components/mermaidInputs/InputWithLabelAndValidation/InputWithLabelAndValidation.js
```diff
@@ -21,13 +35,18 @@ const InputWithLabelAndValidation = ({
...
+  checkboxLabel,
+
   ...restOfProps
 }) => {
   const textFieldRef = useRef()
 
   useStopInputScrollingIncrementNumber(textFieldRef)
 
   const [isHelperTextShowing, setIsHelperTextShowing] = useState(false)
+  const [isChecked, setIsChecked] = useState(false)
```
mmnoo, line 49: Since this is a generic input component (not specific to checkboxes), we could name this state to be `isCheckboxChecked`?

src/components/mermaidInputs/InputWithLabelAndValidation/InputWithLabelAndValidation.js
mmnoo: might be worth adding an inline comment somewhere mentioning that an example use case for the checkbox thing is benthic pit.

src/components/mermaidInputs/InputWithLabelAndValidation/InputWithLabelAndValidation.js
mmnoo: Code philosophy cheap ideas, not a change request: I wonder if another approach is just to wrap this existing input component and add the checkbox to the wrapper (if the checkbox isnt nested). If it is nested, maybe we modify this one to have a slot for the wrapper to inject the checkbox. Thinking in terms of avoiding hasty abstractions, and extension over modification.

mmnoo (approved):
Cutting some corners due to time constraints. Didnt test. Didnt read through requirements. So take with salt. Just review purely from code

src/components/mermaidInputs/InputWithLabelAndValidation/InputWithLabelAndValidation.js
julia-rom: well update the var name ! I'm thinking of adding a separate tasks to make checkboxes more consistent across the site. I noticed we are using different styles, and wrappers.

src/components/mermaidInputs/InputWithLabelAndValidation/InputWithLabelAndValidation.js
julia-rom: good idea! will add it in

PR #912, opened 2024-01-30
[618] make site maps consistent

julia-rom:
[trello ticket](https://trello.com/c/XsNouvbj/618-make-site-maps-consistent)

there are two items in here that need more time: 
- [replace pins on multiple-pin maps](https://trello.com/c/hYzpFwiM/656-update-pins-on-multiple-pin-maps-to-match-single-map-pins) (more context in ticket - wasn’t working as expected with mapbox documentation)
- [prevent re-centering when pin is placed](https://trello.com/c/aIfLnoMV/657-dont-recenter-the-map-when-the-pin-is-placed-by-clicking)

in the mean time,
- pins on single site maps have been replaced
- zoom added to compare duplicate sites
- initial zoom level 13 added to single site map


src/components/mermaidMap/Map.styles.js
```diff
@@ -17,16 +18,15 @@ export const MapWrapper = styled.div`
...
-    height: 44px;
+    width: 31px;
+    height: 31px;
     cursor: move;
   }
 
   svg {
-    width: 3rem;
-    height: 3rem;
+    display: none;
```
mmnoo, line 29: Interesting. Why did we need to supress an svg from being visible. Is that a hack? Is the styling for this SVG set somewhere else and we are undoing it here?

src/components/mermaidMap/ProjectSitesMap/ProjectSitesMap.js
```diff
@@ -43,6 +43,7 @@ const ProjectSitesMap = ({ sitesForMapMarkers, choices }) => {
       customAttribution: language.map.attribution,
     })
 
+    // adds zoom and rotation controls to the map.
     addMapController(map.current)
```
mmnoo, line 47: I prefer changing function names if possible to adding comments in general

src/components/mermaidMap/SingleSiteMap/SingleSiteMap.js
```diff
@@ -48,6 +48,12 @@ const SingleSiteMap = ({
   const [displayHelpText, setDisplayHelpText] = useState(false)
   const [isMarkerBeingPlaced, setIsMarkerBeingPlaced] = useState(false)
 
+  const outOfRangeLatitude = formLatitudeValue > 90 || formLatitudeValue < -90
+
+  const nullishLatitudeOrLongitude =
```
mmnoo, line 53: Always appreciate well named vars for conditionals :)

mmnoo (approved):
Moving quick, so robot tone. Looks good, left some comments

src/components/mermaidMap/ProjectSitesMap/ProjectSitesMap.js
julia-rom: will rename to `addZoomController`

src/components/mermaidMap/Map.styles.js
julia-rom: here we added a custom marker png (via map-libre add marker class) instead of the built-in svg

PR #913, opened 2024-02-01
[633] update pin button states, update change pin button text

julia-rom:
[trello card](https://trello.com/c/ZBgnL5DE/633-replace-pin-button-states)

to test:

1. go to single site map
2. click place site marker
3. ensure value switches to cancel if no new map placement
4. show 'done' if marker is placed
5. show 'replace pin' if there are existing lng/lat values

### DIFF
--- a/src/components/mermaidMap/SingleSiteMap/SingleSiteMap.js
+++ b/src/components/mermaidMap/SingleSiteMap/SingleSiteMap.js
@@ -47,6 +47,10 @@ const SingleSiteMap = ({
   const recordMarker = useRef(null)
   const [displayHelpText, setDisplayHelpText] = useState(false)
   const [isMarkerBeingPlaced, setIsMarkerBeingPlaced] = useState(false)
+  const [placeMarkerButtonText, setPlaceMarkerButtonText] = useState(
+    language.pages.siteForm.placeMarker,
+  )
+  const [hasLatLngChanged, setHasLatLngChanged] = useState(false)
 
   const handleZoomDisplayHelpText = (displayValue) => setDisplayHelpText(displayValue)
   const handleMarkerLocationChange = useCallback(
@@ -62,6 +66,7 @@ const SingleSiteMap = ({
 
       handleLatitudeChange(roundToSixDecimalPlaces(lngLat.lat))
       handleLongitudeChange(roundToSixDecimalPlaces(adjustedLng))
+      setHasLatLngChanged(true)
     },
     [handleLatitudeChange, handleLongitudeChange],
   )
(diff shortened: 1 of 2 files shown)


src/components/mermaidMap/SingleSiteMap/SingleSiteMap.js
```diff
@@ -169,19 +174,34 @@ const SingleSiteMap = ({
...
   const updateBenthicLayers = (dataLayerFromLocalStorage) =>
     setGeomorphicOrBenthicLayerProperty(map.current, 'atlas-benthic', dataLayerFromLocalStorage)
-  const placeMarkerButtonText =
-    formLatitudeValue && formLongitudeValue
-      ? language.pages.siteForm.replaceMarker
-      : language.pages.siteForm.placeMarker
 
   const handlePlaceMarkerClick = () => {
     setIsMarkerBeingPlaced(!isMarkerBeingPlaced)
+    hasLatLngChanged && setHasLatLngChanged(false)
```
mmnoo, line 180: did you mean to wrap this in an if statement. instead of using `&&` (which will work coincidentally). 

You could also just call `setHasLatLngChanged(false)` without checking since internally React wont do anything if its already the value you are setting it to (I believe)

src/components/mermaidMap/SingleSiteMap/SingleSiteMap.js
```diff
@@ -169,19 +174,34 @@ const SingleSiteMap = ({
...
+        setPlaceMarkerButtonText(language.pages.siteForm.placeMarker)
+      }
+      if (!isMarkerBeingPlaced && formLatitudeValue && formLongitudeValue) {
+        setPlaceMarkerButtonText(language.pages.siteForm.replaceMarker)
+      }
+      if (isMarkerBeingPlaced && !hasLatLngChanged) {
+        setPlaceMarkerButtonText(language.pages.siteForm.cancelPlaceMarker)
+      }
+      if (isMarkerBeingPlaced && hasLatLngChanged) {
+        setPlaceMarkerButtonText(language.pages.siteForm.done)
```
mmnoo, line 195: UX question: what if we have changed the location of the marker, but want to cancel? We can only select 'done'.

mmnoo: Is cancel and done doing the same operation here, but we are just changing language for the user? Is that creating low-value complexity in the code?

mmnoo (approved):

mmnoo:
I couldnt test this because the preview link is broken. @saanobhaai?

src/components/mermaidMap/SingleSiteMap/SingleSiteMap.js
julia-rom: good point - maybe we add two buttons , cancel and done

julia-rom: I think it's more for the user - if they activate the pin and want to get out, they can click cancel. If they activate it + move it, it would likely be intentional (when done appears)

src/components/mermaidMap/SingleSiteMap/SingleSiteMap.js
julia-rom: co-pilot recommended this one ... tried it out and it worked - but I think it's clearer if I wrap in an if statement

PR #914, opened 2024-02-02
[632] add long lat swap button to site form

julia-rom:
[trello card](https://trello.com/c/RddqoivR/632-provide-a-button-to-switch-lat-and-long-when-creating-or-updating-a-site-map)

to test:

1. add a new site
2. enter long or lat value
3. press swap button
4. should swap values and update pin on map
5. buttons should be disabled if no lng or lat values (both have to be empty)

### DIFF
--- a/src/components/mermaidInputs/InputWithLabelAndValidation/InputWithLabelAndValidation.js
+++ b/src/components/mermaidInputs/InputWithLabelAndValidation/InputWithLabelAndValidation.js
@@ -14,7 +14,7 @@ import InputNumberNoScrollWithUnit from '../../generic/InputNumberNoScrollWithUn
 
 import InputValidationInfo from '../InputValidationInfo/InputValidationInfo'
 import mermaidInputsPropTypes from '../mermaidInputsPropTypes'
-import { IconButton, CheckBoxLabel } from '../../generic/buttons'
+import { IconButton } from '../../generic/buttons'
 import { IconInfo } from '../../icons'
 
 const InputWithLabelAndValidation = ({
@@ -28,10 +28,9 @@ const InputWithLabelAndValidation = ({
   unit,
   validationMessages,
   validationType,
-  addCheckbox,
-  handleCheckboxUpdate,
-  checkboxLabel,
   renderItemWithinInput,
+  renderItemAboveInput,
+  isInputDisabled,
 
   ...restOfProps
 }) => {
@@ -40,7 +39,6 @@ const InputWithLabelAndValidation = ({
   useStopInputScrollingIncrementNumber(textFieldRef)
 
   const [isHelperTextShowing, setIsHelperTextShowing] = useState(false)
-  const [isCheckboxChecked, setIsCheckboxChecked] = useState(false)
 
   const handleInfoIconClick = (event) => {
     isHelperTextShowing ? setIsHelperTextShowing(false) : setIsHelperTextShowing(true)
@@ -54,7 +52,7 @@ const InputWithLabelAndValidation = ({
       aria-describedby={`aria-descp${id}`}
       id={id}
       unit={unit}
-      disabled={isCheckboxChecked}
+      disabled={isInputDisabled}
       {...restOfProps}
     />
   ) : (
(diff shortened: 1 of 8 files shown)


src/components/mermaidInputs/InputWithLabelAndValidation/InputWithLabelAndValidation.js
```diff
@@ -1,28 +1,21 @@
...
 import InputNumberNoScrollWithUnit from '../../generic/InputNumberNoScrollWithUnit'
 
 import InputValidationInfo from '../InputValidationInfo/InputValidationInfo'
 import mermaidInputsPropTypes from '../mermaidInputsPropTypes'
-import { IconButton } from '../../generic/buttons'
+import { IconButton, CheckBoxLabel, InputButton } from '../../generic/buttons'
 import { IconInfo } from '../../icons'
-import theme from '../../../theme'
-
-const CheckBoxLabel = styled.label`
```
mmnoo, line 15: I dont see this component being used generically. Is it safe to add to the generics or is it specific to this `InputWithLabelAndValidation` use case?

src/components/pages/Site/Site.js
```diff
@@ -114,6 +115,11 @@ const SiteForm = ({
   handleLatitudeChange,
   handleLongitudeChange,
 }) => {
+  const handleLngLatSwap = () => {
+    handleLatitudeChange(formik.getFieldProps('longitude').value)
+    handleLongitudeChange(formik.getFieldProps('latitude').value)
```
mmnoo, line 120: should we store one or both of these in an intermediate value in case latitude gets updated so fast on line 120 when the code reads the value from latitude, its the new longitude value?

src/components/pages/Site/Site.js
```diff
@@ -165,6 +171,11 @@ const SiteForm = ({
           helperText={language.helperText.getLatitude()}
           shouldShowSteps={true}
           step="0.000001"
+          addInputButton={true}
```
mmnoo, line 174: I wonder if we just provide a render slot for anything to take up the space that the inputButton would. Since this abstraction is becoming more and more hairy. That would make this abstraction more flexible?

src/components/mermaidInputs/InputWithLabelAndValidation/InputWithLabelAndValidation.js
```diff
@@ -102,10 +100,24 @@ const InputWithLabelAndValidation = ({
               checked={isCheckboxChecked}
               onChange={(event) => handleCheckboxChange(event.target.checked)}
             />
+
             {checkboxLabel}
           </CheckBoxLabel>
         ) : null}
-        {inputType}
+        <InputContainer>
```
mmnoo, line 107: following up on my other comment. I wonder if here we just have a render slot and we can add any component to (and the logic for that component can live elsewhere. I cant see how this looks, but we could name that render slot based on its UI position. This adds theoretical flexibility, but also helps keep that logic defined elsewhere so that this generic component doesnt become too bloated.

eg `<InputWithLabelAndValidation leftItem={<AnyComponet doThis={x} isFoo={false}/>}`

mmnoo (commented):
Im not sure about the extra stuff added to the `InputLableAndValidation` component. Depending on your balance of tradeoffs regarding time with this ticket, we might consider making it be a dumber component and use composition via render slots to keep extra logic external to it.

src/components/mermaidInputs/InputWithLabelAndValidation/InputWithLabelAndValidation.js
julia-rom: it's only being used in one instance now - but recently we've been adding extras to inputs (checkboxes + buttons). I sense there will be more cases like this added in the future

src/components/mermaidInputs/InputWithLabelAndValidation/InputWithLabelAndValidation.js
julia-rom: I need to think more on this one , as there are two things being rendered in different spots (checkbox and button) ... good to bring up though!

julia-rom: cleaned this up - thanks for the feedback! removed checkbox and button from this component and just passed both as rendered items (within or above input)

src/components/pages/Site/Site.js
julia-rom: I'm a bit confused - `formik.getFieldProps('longitude').value` is the most recent value. We're use similarly on `ProjectInfo.js` and `ManagemenetRegime.js`

src/components/pages/Site/Site.js
julia-rom: done and done ✅

PR #916, opened 2024-02-28
[632] bug fix for lng/lat swap errors

julia-rom:
[ticket here](https://trello.com/c/RddqoivR/632-provide-a-button-to-switch-lat-and-long-when-creating-or-updating-a-site-map)

Tiela added a video in the ticket on how to test

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved latitude and longitude swap functionality in the site form to ensure correct value placement and enhanced form validation.
	- Added a useEffect hook to trigger form validation after updating form values.

### DIFF
--- a/src/components/pages/Site/Site.js
+++ b/src/components/pages/Site/Site.js
@@ -140,7 +140,7 @@ const SiteForm = ({
   useEffect(() => {
     formik.validateForm()
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [formik.getFieldProps('latitude').value, formik.getFieldProps('longitude').value])
+  }, [formik.values.latitude, formik.values.longitude])
 
   return (
     <form id="site-form" onSubmit={formik.handleSubmit}>


src/components/pages/Site/Site.js
```diff
@@ -126,10 +126,22 @@ const SiteForm = ({
...
+    const currentLongitude = formik.getFieldProps('longitude').value
+
+    handleLatitudeChange(currentLongitude)
+    handleLongitudeChange(currentLatitude)
+
+    formik.setFieldTouched('longitude', true)
+    formik.setFieldTouched('latitude', true)
   }
 
+  // hack to ensure formik values are updated before manually triggering validation
```
mmnoo, line 139: instead of a comment, can we name the effect? Either `const _someExplaningName = useEffect...` or `useEffect(function someExplainingName() {}, []}`

Not a blocker, but if we can avoid a comment, we should IMO.

src/components/pages/Site/Site.js
```diff
@@ -126,10 +126,22 @@ const SiteForm = ({
...
+  useEffect(() => {
+    formik.validateForm()
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [formik.getFieldProps('latitude').value, formik.getFieldProps('longitude').value])
```
mmnoo, line 143: Im not sure about the deps array. Not that there is a lint exception, but that we are calling a function as a dep array value. Im not sure how that will work interms of how the effect triggers. Might be safer with `formik.values.<fieldName>`

mmnoo (commented):
Conditional approval on calling a function in the deps array being changed to a value. 

Also if were making code changes anyway, removing the comment in favour of self documenting code is recommended.

src/components/pages/Site/Site.js
julia-rom: oh good idea - well swap for `formik.values.<fieldName>` and test it!

src/components/pages/Site/Site.js
julia-rom: will update to a named effect

PR #961, opened 2024-05-17
Feat: F4 updates

JBurkinshaw:
**Summary**:

- Update F4 null value logic.
- Remove button under the inputs.
- Disable the “save and update” button if no changes will be made when the user clicks it. Update language.
- Fix issue when 0.0 displayed when values set to 0.

Refs: https://trello.com/c/djRMhpDu/789-updates-to-f4, https://trello.com/c/Y8IhoOuM/788-f4-null-values-from-mermaid

## Summary by CodeRabbit

- **New Features**
  - Added dynamic helper text properties to various form fields for improved guidance.
  - Introduced a new function to calculate the total of two field values.

- **Enhancements**
  - Improved language variable handling for better localization in forms.
  - Enhanced logic to disable "save and update" button based on conditions.

- **Bug Fixes**
  - Updated input blur handling to correctly round values to one decimal place if they are not integers.


alanjleonard (changes requested):
- Changes to the inputs OR the date field should re-enable the button, since that would calculate different values for the inputs. The first thing the user will likely do is change the date range.
- Clicking the button should always override the values then disable the button
- Right now if you change a value, then click the button, it only updates the helper text. Then if you click it again, it updates the value. Clicking the button should update the values to what's been calculated for that date range and disable the button.

alanjleonard (changes requested):
Looks good. Don't show the warning toast if we can't get a value for any of the F4s. It may be interpreted as an error. The helper text covers the explanation.

PR #977, opened 2024-07-09
[795] image classification: sample unit selector

julia-rom:
[trello card](https://trello.com/c/7bAtoXXm/795-sample-unit-input-selector)

the beginning template for the sample unit selector - not sure what we are adding in the description yet, so lorum ipsum text is there

Screenshot 2024-07-09 at 11 22 40 AM

### DIFF
--- a/src/components/pages/ImageClassification/ImageClassificationTable.js
+++ b/src/components/pages/ImageClassification/ImageClassificationTable.js
@@ -1,10 +1,10 @@
 import React from 'react'
-import PropTypes from 'prop-types'
+// import PropTypes from 'prop-types'
 
 const ImageClassiciationTable = () => {
   return <p>IMAGE CLASSICIATION TABLE</p>
 }
 
-ImageClassiciationTable.propTypes = {}
+// ImageClassiciationTable.propTypes = {}
 
 export default ImageClassiciationTable
--- a/src/components/pages/ImageClassification/ImageClassificationTable.js
+++ b/src/components/pages/ImageClassification/ImageClassificationTable.js
@@ -2,9 +2,9 @@ import React from 'react'
 // import PropTypes from 'prop-types'
 
 const ImageClassificationTable = () => {
-  return <p>IMAGE CLASSICIATION TABLE</p>
+  return <p>IMAGE CLASSIFICATION TABLE</p>
 }
 
-// ImageClassiciationTable.propTypes = {}
+// ImageClassificationTable.propTypes = {}
 
 export default ImageClassificationTable
--- a/src/components/pages/ImageClassification/ImageClassificationTable.js
+++ b/src/components/pages/ImageClassification/ImageClassificationTable.js
@@ -1,0 +1,10 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+const ImageClassiciationTable = () => {
+  return <p>IMAGE CLASSICIATION TABLE</p>
+}
+
+ImageClassiciationTable.propTypes = {}
+
+export default ImageClassiciationTable
--- a/src/components/pages/ImageClassification/ImageClassificationTable.js
+++ b/src/components/pages/ImageClassification/ImageClassificationTable.js
@@ -1,10 +1,10 @@
 import React from 'react'
 // import PropTypes from 'prop-types'
 
-const ImageClassiciationTable = () => {
+const ImageClassificationTable = () => {
   return <p>IMAGE CLASSICIATION TABLE</p>
 }
 
 // ImageClassiciationTable.propTypes = {}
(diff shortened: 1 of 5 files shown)


src/components/pages/ImageClassification/ImageClassificationTable.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react'
+// import PropTypes from 'prop-types'
+
+const ImageClassiciationTable = () => {
```
julia-rom, line 4: this will be added in the next task

julia-rom: the typo will also be fixed

alanjleonard (changes requested):
1. This should only be on new BPQ sample units. If there's already manual observations, just show the observation table.
2. From `theme.js` Use `grey5` as the background for the callout, and use `maxTextWidth` for the `max-width` of the paragraph.

src/components/pages/ImageClassification/ImageClassificationTable.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react'
+// import PropTypes from 'prop-types'
+
+const ImageClassificationTable = () => {
```
julia-rom, line 4: coming in the next card ...

src/language.js
```diff
@@ -961,6 +961,16 @@ const tooltipText = {
   softCoralPercentage: 'Soft coral cover as decimal percentage of quadrat total area (e.g. 33.3).',
 }
 
+const imageClassification = {
+  sampleUnitInputSelector: {
+    title: 'Image classification is now available in MERMAID.',
+    description:
+      'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Phasellus vel leo pellentesque lorem interdum pretium. Aenean varius luctus ex, vel blandit elit mollis non. Nam risus felis, finibus et nisl vel, convallis viverra odio. Nunc ut turpis tortor. Praesent consectetur nibh quis tortor bibendum fermentum',
```
julia-rom, line 968: still being figured out ... leaving as lorem ipsum for now

alanjleonard (approved):

src/components/pages/collectRecordFormPages/BenthicPhotoQuadratForm/BenthicPhotoQuadratForm.js
```diff
@@ -208,22 +211,39 @@ const BenthicPhotoQuadratForm = ({ isNewRecord = true }) => {
...
+      } else if (observationTableType === 'image-classification') {
         return (
-          <BenthicPhotoQuadratObservationTable
-            benthicAttributeSelectOptions={benthicAttributeSelectOptions}
-            {...props}
-          />
+          <>
+            <ImageClassificationTable />
+          </>
         )
```
gwan93, line 236: The fragments aren't necessary here since you're only returning a single element in both cases

julia-rom: good point! will remove

src/components/pages/ImageClassification/SampleUnitInputSelector.js
```diff
@@ -0,0 +1,62 @@
...
+const SampleUnitInputSelector = ({ setObservationTableType }) => {
+  const handleSampleUnitChange = (type) => () => {
+    setObservationTableType(type)
+  }
+
+  return (
+    <SelectorContainer>
+      <TextContainer>
+        <H3 htmlFor="image-classification-selection">
+          {language.imageClassification.sampleUnitInputSelector.title}
```
JBurkinshaw, line 38: You could potential destructure the relevant object in the language file to make this less verbose.
e.g.
```const { sampleUnitInputSelector: sampleUnitInputSelectorLanguage } = language.imageClassification```
or
```const { title, description, etc} = language.imageClassification.sampleUnitInputSelector```

I did something similar in the GFCR components: https://github.com/data-mermaid/mermaid-webapp/blob/d0a33c14bf8b8122d8ff6483f5db6c7db284486f/src/components/pages/gfcrPages/GfcrIndicatorSetForm/subPages/F1Form.js#L11

JBurkinshaw (commented):
Sorry about the delay on this. Looks good to me. One minor suggestion but no need to change it now it's merged.

src/components/pages/ImageClassification/SampleUnitInputSelector.js
julia-rom: good idea - will update this in the next branch and also use this in other components 🙌

PR #981, opened 2024-07-11
Fix PR Previews

parmvirthind:
- Added `basename` property to `createBrowserRouter` as it was omitted when we upgraded from react router v5 to v6.
  - This property tells the application the starting point and routes accordingly. Its value should match the `homepage` value in `package.json`.
  - This is needed for PR previews where the `homepage` value is set to the PR Number. (ex. `https://preview.app2.datamermaid.org/<PR_NUMBER>/index.html`)

## Summary by CodeRabbit

- **New Features**
  - Improved routing configuration to better handle base URLs, ensuring more consistent navigation within the app.


julia-rom (approved):
nice!!

PR #983, opened 2024-07-15
M916 cpu spike

parmvirthind:
[Relates to Ticket 916](https://trello.com/c/55knHtKP/916-cpu-spike-on-sites-page)

Changes:
- Updated `addClusterSourceAndLayers` to initialize with an empty array of features instead of `sitesForMapMarkers`.
  - This is because we have a `useEffect` that updates the source data based on the value of `sitesForMapMarkers`.
- Updated the dependencies for the `_initializeMap` `useEffect`. 
  - Previously it was using `sitesForMapMarkers` as a dependency, this was causing the entire map to re-render whenever the user would submit a query. 
  - Now, the `useEffect` only runs when our page loads, and the `_updateMapMarkers` `useEffect` handles the query.
- Removed the `sourcedata` event.
  - This was causing the infinite re-render as explained in the ticket linked above. Instead we handle updating the markers through a `useEffect` that is dependent on `sitesForMapMarkers`

## Summary by CodeRabbit

- **Refactor**
  - Streamlined map initialization process and marker updates for improved performance.
  - Removed unused `usePrevious` hook.
  - Simplified internal logic of functions to reduce code complexity.

### DIFF
--- a/src/components/mermaidMap/ProjectSitesMap/ProjectSitesMap.js
+++ b/src/components/mermaidMap/ProjectSitesMap/ProjectSitesMap.js
@@ -43,17 +43,17 @@ const ProjectSitesMap = ({ sitesForMapMarkers, choices }) => {
     addZoomController(map.current)
 
     map.current.on('load', () => {
-      addClusterSourceAndLayers(map.current, sitesForMapMarkers)
+      addClusterSourceAndLayers(map.current)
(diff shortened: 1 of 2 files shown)


julia-rom (approved):
looks good - I didn't realize this was happening before. Just tested and feels smooth 🙌

PR #987, opened 2024-07-17
Feat: Use required dropdown instead of checkbox

JBurkinshaw:
Applied to Finance Solution and Revenue Stream modals.

Ref: https://trello.com/c/j4ywl8rK/853-use-a-required-dropdown-for-checkboxes-with-yes-no-options

## Summary by CodeRabbit

- **New Features**
    - Added new form fields for `local_enterprise` and `gender_smart` in the Finance Solution Modal.

- **Enhancements**
    - Updated the `sustainable_revenue_stream` field in Revenue Modal to use a `Select` component with 'Yes' and 'No' options for improved user experience.

- **Bug Fixes**
    - Initial values for `gender_smart` and `local_enterprise` are now set to empty strings instead of `false` to ensure proper form handling.

- **Localization**
    - Added 'yes' and 'no' translations to support new form fields.

### DIFF
--- a/src/components/pages/gfcrPages/GfcrIndicatorSetForm/modals/FinanceSolutionModal.js
+++ b/src/components/pages/gfcrPages/GfcrIndicatorSetForm/modals/FinanceSolutionModal.js
@@ -142,7 +142,7 @@ const FinanceSolutionModal = ({
         errors.local_enterprise = [{ code: language.error.formValidation.required, id: 'Required' }]
       }
 
-      if (!values.gender_smart === '') {
+      if (values.gender_smart === '') {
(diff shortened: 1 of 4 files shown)


alanjleonard (changes requested):
If you change Gender 2x Criteria back to "choose..." the "Add finance solution row" button stays enabled. It should disable if it's not set to "yes" or "no".

JBurkinshaw:
> If you change Gender 2x Criteria back to "choose..." the "Add finance solution row" button stays enabled. It should disable if it's not set to "yes" or "no".

Well spotted. There was a bug in the validation logic. Updated

PR #989, opened 2024-07-18
fix broken tests: add mock cookie beforeEach

julia-rom:
this is on an existing branch `shared-logins-with-dashboard` which originally broke tests due to a missing mock cookie document. The fix consisted of adding this in `beforeEach` test in `setupTests.js`

to test:

1. pull this branch locally
2. run `yarn test a`
3.  ensure all tests pass

going forward, I think it's a good idea to add this additional check on all pr's, incase the original dev forgets to do it

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved `useAuthentication` function to prevent errors in non-browser environments by checking for the existence of `document`.

- **Tests**
  - Enhanced test setup with `mockDocumentCookie` to simulate authentication-related cookies.


src/App/useAuthentication.js
```diff
@@ -48,12 +48,15 @@ const useAuthentication = ({ dexieCurrentUserInstance }) => {
     let isMounted = true
     const auth0CookieName =
       ' ' + `auth0.${process.env.REACT_APP_AUTH0_CLIENT_ID}.is.authenticated=true`
-    const auth0CookieExists = document.cookie.split(';').includes(auth0CookieName)
+
+    const auth0CookieExists =
+      typeof document !== 'undefined' && document.cookie.split(';').includes(auth0CookieName)
```
julia-rom, line 53: add check to ensure tests don't break due to missing document

julia-rom: that works too - will update

PR #992, opened 2024-07-19
Feat: Switch to checkboxes for the sustainable finance mechanism multi-select

JBurkinshaw:
Ref: https://trello.com/c/ac3qvqCU/847-use-checkboxes-for-the-sustainable-finance-mechanism-multi-select

## Summary by CodeRabbit

- **New Features**
  - Enhanced multi-selection functionality in the finance solution modal with checkboxes next to each option, allowing users to select multiple options simultaneously.


alanjleonard (changes requested):
Use the default font (colour, family, size) for the checkbox options.

PR #1002, opened 2024-08-01
M946: Image Annotation Modal (image): Display Patches on Image

parmvirthind:
Relates to [Ticket 946](https://trello.com/c/SOW9zFSL/946-image-annotation-modal-image-display-patches-on-map)

**To test:**
- Open Modal
- Ensure that there is a grid of boxes overlaying the image
  - Scaling of the patches/points/boxes will be handled in M949
- Ensure that you can zoom in and out and the dimensions of the grid stays the same


julia-rom (approved):
good you have some sample data to work with!

PR #1006, opened 2024-08-07
M949: Image Annotation Modal (image): Image Scaling

parmvirthind:
Relates to [Ticket 949](https://trello.com/c/CJpniAjC/949-image-annotation-modal-image-image-scaling)

**Changes/Notes:**
- Added `sample-data.js` to mock data returned from Image Response API call
- `PATCH_SIZE` is hardcoded for now. It sounds like it will be a part of the API response in the future but not immediately.
- `MAX_DIMENSION` will likely be changed in the future , but it is unclear how it will need to be implemented until mockup is finalized.
  - This value might by dynamic based on screen-size or it might stay a static value
  - It also assumes that the height and width have the same max dimension, which may not be the case
- Instead of hardcoding Map dimensions, we now retrieve those values from the uploaded image
- Implemented image scaling to account for likely large images (ex. 8000x8000)
- Simplified state/props to only need `[dataToReview, setDataToReview]`
  - This prevents the need from managing `isModalOpen` state which can be inferred by `dataToReview` state.
-  Updated math for patches to account for image scaling

**To Test:**
1. Go to benthic photo quadrat and select "Use image classification..."
2. Upload any photo or file, it will not matter
3. Click "Review" on uploaded file row
4. Ensure that the test image is displayed with proper dimensions and that the yellow patches align with the first two rows of red patches from the image. 

**Screenshots:**
Screenshot 2024-08-07 at 12 27 53 PM

### DIFF
--- a/src/components/pages/ImageClassification/ImageAnnotationModal.js
+++ b/src/components/pages/ImageClassification/ImageAnnotationModal.js
@@ -2,6 +2,7 @@ import React, { useEffect, useRef, useState } from 'react'
 import PropTypes from 'prop-types'
 import styled from 'styled-components'
 import maplibregl from 'maplibre-gl'
+import { toast } from 'react-toastify'
 import Modal from '../../generic/Modal/Modal'
 
 // TODO: In future, PATCH_SIZE will come from API
@@ -50,6 +51,11 @@ const ImageAnnotationModal = ({ dataToReview, setDataToReview }) => {
         width: imageForMap.naturalWidth * imageScale,
         scale: imageScale,
       })
+    }
+
+    imageForMap.onerror = () => {
+      setDataToReview()
+      toast.error('There was a problem displaying this image, please contact if issue persists.')
     }
     // eslint-disable-next-line
   }, [])


src/components/pages/ImageClassification/ImageAnnotationModal.js
```diff
@@ -82,12 +33,29 @@ const LegendSquare = styled.div`
...
+      const imageScale = longerSide > MAX_DIMENSION ? MAX_DIMENSION / longerSide : 1
+
+      setDimensions({
+        height: imageForMap.naturalHeight * imageScale,
+        width: imageForMap.naturalWidth * imageScale,
+        scale: imageScale,
+      })
+    }
+    // eslint-disable-next-line
+  }, [])
```
julia-rom, line 55: don't we need `dataToReview.image` as a dep?

julia-rom: could add error handling `imageForMap.onerror`

julia-rom (commented):
looks like a dep is missing + image upload could use error handling in the useEffect

src/components/pages/ImageClassification/ImageAnnotationModal.js
julia-rom: yes i think closing the modal + toast is a good idea - and same for what i've implemented in for the image drop modal

PR #1014, opened 2024-08-16
M961 delete row

parmvirthind:
Relates to [Ticket 961](https://trello.com/c/vgWEYbhv/961-image-annotation-modal-table-delete-row)

**Changes:**
- added styling for unclassified points
- re-factored `ImageAnnotationModalMap` component
  - now that we are receiving image dimensions from the API response, we no longer need to calculate it before setting up the map
  - put geojson for points into a `useCallback` so that it can be re-used
- added functionality where the user can delete a row in the table, and the styling for points on the map will be updated. 

**To Test:**
1. do the whole image upload thing and then click"Review" button for an image
2. Remove a row from the table by clicking the red "X" button in the last column
3. Note that the row is removed from the table, and that the points corresponding to that row are now the unclassified colour (red)

### DIFF
--- a/src/components/pages/ImageClassification/ImageAnnotationModal/ImageAnnotationModalTable.js
+++ b/src/components/pages/ImageClassification/ImageAnnotationModal/ImageAnnotationModalTable.js
@@ -48,6 +48,7 @@ const ImageAnnotationModalTable = ({
     setDataToReview((prevState) => ({ ...prevState, points: updatedPoints }))
     setSelectedRowIndex()
     setSelectedPoints([])
+    setHighlightedPoints([])
   }
 
   return (


src/components/pages/ImageClassification/ImageAnnotationModal/ImageAnnotationModalMap.js
```diff
@@ -16,44 +15,77 @@ const SELECTED_POLYGON_LINE_WIDTH = 10
...
-  const _getImageDimensions = useEffect(() => {
-    const imageForMap = new Image()
-    imageForMap.src = dataToReview.image
-    imageForMap.onload = () => {
-      const longerSide = Math.max(imageForMap.naturalWidth, imageForMap.naturalHeight)
-      const imageScale = longerSide > MAX_DIMENSION ? MAX_DIMENSION / longerSide : 1
-
-      setDimensions({
-        height: imageForMap.naturalHeight * imageScale,
-        width: imageForMap.naturalWidth * imageScale,
```
julia-rom, line 43: could add `PATCH_SIZE / 2 `as a var

src/components/pages/ImageClassification/ImageAnnotationModal/ImageAnnotationModalMap.js
```diff
@@ -16,44 +15,77 @@ const SELECTED_POLYGON_LINE_WIDTH = 10
...
-      toast.error('There was a problem displaying this image, please contact if issue persists.')
-    }
-    // eslint-disable-next-line
-  }, [])
+  const [hasMapLoaded, setHasMapLoaded] = useState(false)
+  const imageScale = getImageScale(dataToReview)
+
+  const getPointsGeojson = useCallback(
+    () => ({
+      type: 'FeatureCollection',
```
julia-rom, line 41: would suggest adding a check: `if (map.currrent)` at the top of this function or `if (!map.current) return`

julia-rom (approved):
could looks good, couple minor suggestions

PR #1019, opened 2024-08-26
M991 display point modal

parmvirthind:
Relates to [Ticket 991](https://trello.com/c/pSBTR8Xe)

**Changes**
- Added a click handler on map points that opens a modal
- Details of this modal are going to be implemented later on when there is more information

### DIFF
--- a/src/components/pages/ImageClassification/ImageAnnotationModal/ImageAnnotationModalMap.js
+++ b/src/components/pages/ImageClassification/ImageAnnotationModal/ImageAnnotationModalMap.js
@@ -7,6 +7,7 @@ import {
   imageClassificationResponsePropType,
 } from '../../../../App/mermaidData/mermaidDataProptypes'
 import Modal from '../../../generic/Modal/Modal'
+import { ImageAnnotationMapContainer } from './ImageAnnotationModal.styles'
 
 // TODO: Assumes that the max dimension for height and width are the same.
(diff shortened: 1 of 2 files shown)


src/components/pages/ImageClassification/ImageAnnotationModal/ImageAnnotationModalMap.js
```diff
@@ -238,14 +245,26 @@ const ImageAnnotationModalMap = ({
...
-    <div
-      style={{
-        width: dataToReview.original_image_width * imageScale,
-        height: dataToReview.original_image_height * imageScale,
-        marginTop: '2rem',
-      }}
-      ref={mapContainer}
-    />
+    <>
+      <div
```
julia-rom, line 249: could add modal style as a const vs inline

julia-rom (commented):
prefer to avoid using inline styling

src/components/pages/ImageClassification/ImageAnnotationModal/ImageAnnotationModalMap.js
julia-rom: hard to see properly in the editor but it's this bit:

Screenshot 2024-08-26 at 3 22 34 PM

julia-rom (approved):

PR #1036, opened 2024-09-11
M999 save point annotation

parmvirthind:
Relates to [Ticket 999](https://trello.com/c/G9Y02Fl3/999-image-annotation-modal-edit-point-update-data-state-on-click-away)

**Changes:**
- After fetching data from API, sort data so that `is_confirmed: true` annotations are moved to the beginning of the annotations array, and then set state
- Fixed issue where points would be re-drawn incorrectly (see `hackTemporarilySetMapToDefaultPosition` in `ImageAnnotationModalMap`). There may be a better solution for this, but with limited time this was the quickest solution. 
- Changed `ImageAnnotationPopup` from a maplibre popup to a react element that is positioned on top-right of map. Was having issue with react state syncing. [I have opened a ticket to investigate this further. ](https://trello.com/c/BWG1pzsO/1021-edit-point-popup-use-maplibre-popup-or-absolute-positioned-container)
- Zoom into point once clicked
- Re-factored `ImageAnnotationModalPopup` into smaller components: `ClassifierGuesses`, `ExistingRows`, `NewRow`
- `ClassifierGuesses`
  - Displays the `machine_generated` annotations from the API, sorted by confidence
  - When a new classifier guess is selected:
    - The selected annotation gets moved to the front of the array
    - The selected annotation gets a `is_confirmed` status of `true`
    - The rest of the annotations are updated to have a `is_confirmed` status of `false`
    - React state for image data is updated.
- `ExistingRows`
  - Options for the dropdown are retrieved by a reducer that runs on the points in which we filter out:
    - unclassified points
    - annotations that are a classifier guess of the selected point
    - annotations that have already been added via the reducer
  - When an option from the dropdown for Existing Rows is selected, we update the annotations for that point so that:
    - We only have classifier guesses (`is_machine_created = true`), and all classifier guesses get a `is_confirmed` status of `false`
    - The selected option is moved to the top of the annotations array with a `is_confirmed` status of `true`
    - React state for image data is updated.
- `NewRow`
  - Ignore this file. Mockup was updated, so there will be new functionality.

**To test:**
- Open the image annotation modal by clicking "Review" on an uploaded image
- click on a point in the map
- select an option other than the first classifier guess (this is already selected, radio button should be checked but isn't which[ will be done in this ticket](https://trello.com/c/fqiXLqC4/1017-edit-point-popup-default-value-for-radio-buttons))
- close the modal
- either inspect the response to ensure that the point was updated, or re-open the modal by clicking "Review" on the same image, and see that the change was persisted. 
- Note: need to disable the close button and display loading indicator while `PATCH` is being made

### DIFF
--- a/src/components/pages/ImageClassification/ImageAnnotationModal/ImageAnnotationModal.js
+++ b/src/components/pages/ImageClassification/ImageAnnotationModal/ImageAnnotationModal.js
@@ -23,15 +23,7 @@ import LoadingIndicator from '../../../LoadingIndicator/LoadingIndicator'
 // Annotations that are "confirmed" get top priority, with the other annotations being sorted by score from the API.
 // When a user selects a new annotation for that point, it is pushed to the first position in the annotations array.
 
-const prioritizeConfirmedAnnotations = (a, b) => {
-  if (a.is_confirmed && !b.is_confirmed) {
-    return -1
-  }
-  if (!a.is_confirmed && b.is_confirmed) {
-    return 1
-  }
-  return 0
-}
+const prioritizeConfirmedAnnotations = (a, b) => b.is_confirmed - a.is_confirmed
 
 const ImageAnnotationModal = ({ imageId, setImageId }) => {
   const { databaseSwitchboardInstance } = useDatabaseSwitchboardInstance()
--- a/src/components/pages/ImageClassification/ImageAnnotationModal/ImageAnnotationPopup/ClassifierGuesses.js
+++ b/src/components/pages/ImageClassification/ImageAnnotationModal/ImageAnnotationPopup/ClassifierGuesses.js
@@ -9,16 +9,12 @@ import {
 
 const moveAnnotationToFront = (array, index) => {
   const newArray = [...array]
-  newArray.unshift(newArray.splice(index, 1)[0])
-  return newArray
+  const [movedAnnotation] = newArray.splice(index, 1)
+  return [movedAnnotation, ...newArray]
 }
 
 const confirmFirstAnnotationAndUnconfirmRest = (annotation, i) => {
-  if (i === 0) {
-    annotation.is_confirmed = true
-  } else {
-    annotation.is_confirmed = false
-  }
+  annotation.is_confirmed = i === 0
 }
 
 const ClassifierGuesses = ({
--- a/src/components/pages/ImageClassification/ImageAnnotationModal/ImageAnnotationPopup/ExistingRows.js
+++ b/src/components/pages/ImageClassification/ImageAnnotationModal/ImageAnnotationPopup/ExistingRows.js
@@ -14,12 +14,12 @@ const isClassified = ({ is_unclassified, annotations }) =>
 const isAClassifierGuessOfSelectedPoint = (annotations, benthic_attribute, growth_form) =>
   annotations.some(
     (annotation) =>
-      !!annotation.is_machine_created &&
+      annotation.is_machine_created &&
       annotation.benthic_attribute === benthic_attribute &&
       annotation.growth_form === growth_form,
   )
 
-const isAlreadyPushed = (acc, value) => acc.some((option) => option.value === value)
+const isOptionAlreadyAdded = (acc, value) => acc.some((option) => option.value === value)
 
 const ExistingRows = ({
   selectedPoint,
@@ -39,7 +39,7 @@ const ExistingRows = ({
 
     if (
       isClassified(currentPoint) &&
-      !isAlreadyPushed(acc, value) &&
+      !isOptionAlreadyAdded(acc, value) &&
       !isAClassifierGuessOfSelectedPoint(selectedPoint.annotations, benthic_attribute, growth_form)
     ) {
       acc.push({ label, value })
@@ -63,8 +63,7 @@ const ExistingRows = ({
     // Only want classifier guesses, and want to them unconfirmed.
     const resetAnnotationsForPoint = selectedPoint.annotations.reduce((acc, currentAnnotation) => {
       if (currentAnnotation.is_machine_created) {
-        currentAnnotation.is_confirmed = false
-        acc.push(currentAnnotation)
+        acc.push({ ...currentAnnotation, is_confirmed: false })
       }
 
       return acc
@@ -74,6 +73,7 @@ const ExistingRows = ({
       benthic_attribute,
       growth_form: growth_form === 'null' ? null : growth_form,
       is_confirmed: true,
+      is_machine_created: false,
     }
 
     const updatedAnnotations = [annotationToAdd, ...resetAnnotationsForPoint]


src/components/pages/ImageClassification/ImageAnnotationModal/ImageAnnotationModal.js
```diff
@@ -16,6 +17,22 @@ import {
 import { useDatabaseSwitchboardInstance } from '../../../../App/mermaidData/databaseSwitchboard/DatabaseSwitchboardContext'
 import LoadingIndicator from '../../../LoadingIndicator/LoadingIndicator'
 
+// Context: Strategy for the Image Annotation Modal is to use the 1st annotation in the array for each point
+// as the "current" annotation. The 1st annotation is used to group points into rows for the table,
+// and also used as the properties for the points on the map.
+// Annotations that are "confirmed" get top priority, with the other annotations being sorted by score from the API.
+// When a user selects a new annotation for that point, it is pushed to the first position in the annotations array.
```
julia-rom, line 24: makes sense!

src/components/pages/ImageClassification/ImageAnnotationModal/ImageAnnotationPopup/ClassifierGuesses.js
```diff
@@ -0,0 +1,87 @@
...
+import PropTypes from 'prop-types'
+import { Tr } from '../../../../generic/Table/table'
+import { PopupTd, PopupTdForRadio } from '../ImageAnnotationModal.styles'
+import {
+  imageClassificationPointPropType,
+  imageClassificationResponsePropType,
+} from '../../../../../App/mermaidData/mermaidDataProptypes'
+
+const moveAnnotationToFront = (array, index) => {
+  const newArray = [...array]
```
julia-rom, line 11: could also do 
```
const moveAnnotationToFront = (array, index) => {
  const [movedAnnotation] = array.splice(index, 1)
  return [movedAnnotation, ...array]
}
```

src/components/pages/ImageClassification/ImageAnnotationModal/ImageAnnotationPopup/ClassifierGuesses.js
```diff
@@ -0,0 +1,87 @@
...
+  newArray.unshift(newArray.splice(index, 1)[0])
+  return newArray
+}
+
+const confirmFirstAnnotationAndUnconfirmRest = (annotation, i) => {
```
julia-rom, line 16: could also name it to `toggleAnnotationConfirmation` for better clarity

src/components/pages/ImageClassification/ImageAnnotationModal/ImageAnnotationPopup/ClassifierGuesses.js
```diff
@@ -0,0 +1,87 @@
...
+  if (i === 0) {
+    annotation.is_confirmed = true
```
julia-rom, line 18: slightly shorter:   `annotation.is_confirmed = index === 0`

src/components/pages/ImageClassification/ImageAnnotationModal/ImageAnnotationPopup/ClassifierGuesses.js
```diff
@@ -0,0 +1,87 @@
...
+  setDataToReview,
+  selectedRadioOption,
+  setSelectedRadioOption,
+  getBenthicAttributeLabel,
+  getGrowthFormLabel,
+}) => {
+  const classifierGuesses = selectedPoint.annotations.filter(
+    (annotation) => annotation.is_machine_created,
+  )
+  const classifierGuessesSortedByScore = classifierGuesses.toSorted((a, b) => b.score - a.score)
```
julia-rom, line 36: if this gets large, might want to memoize it

src/components/pages/ImageClassification/ImageAnnotationModal/ImageAnnotationPopup/ExistingRows.js
```diff
@@ -0,0 +1,135 @@
...
+  imageClassificationResponsePropType,
+} from '../../../../../App/mermaidData/mermaidDataProptypes'
+
+const isClassified = ({ is_unclassified, annotations }) =>
+  !is_unclassified && annotations.length > 0
+
+const isAClassifierGuessOfSelectedPoint = (annotations, benthic_attribute, growth_form) =>
+  annotations.some(
+    (annotation) =>
+      !!annotation.is_machine_created &&
```
julia-rom, line 17: this should already be a boolean

src/components/pages/ImageClassification/ImageAnnotationModal/ImageAnnotationPopup/ExistingRows.js
```diff
@@ -0,0 +1,135 @@
...
+      annotation.benthic_attribute === benthic_attribute &&
+      annotation.growth_form === growth_form,
+  )
+
+const isAlreadyPushed = (acc, value) => acc.some((option) => option.value === value)
```
julia-rom, line 22: name suggestion: `isOptionAlreadyAdded` for a bit more clarity

src/components/pages/ImageClassification/ImageAnnotationModal/ImageAnnotationPopup/ExistingRows.js
```diff
@@ -0,0 +1,135 @@
...
+      selectedPoint.annotations[0].growth_form
+    const isPointInARow = existingRowDropdownOptions.some((row) => rowKeyForPoint === row.value)
+    return isPointInARow ? rowKeyForPoint : ''
+  })
+
+  const addExistingAnnotation = (existingAnnotation) => {
+    const [benthic_attribute, growth_form] = existingAnnotation.split('_')
+
+    // Only want classifier guesses, and want to them unconfirmed.
+    const resetAnnotationsForPoint = selectedPoint.annotations.reduce((acc, currentAnnotation) => {
```
julia-rom, line 64: would prefer to not directly mutate the `currentAnnotation` object  and make a copy instead

src/components/pages/ImageClassification/ImageAnnotationModal/ImageAnnotationPopup/ExistingRows.js
```diff
@@ -0,0 +1,135 @@
...
+        currentAnnotation.is_confirmed = false
+        acc.push(currentAnnotation)
+      }
+
+      return acc
+    }, [])
+
+    const annotationToAdd = {
+      benthic_attribute,
+      growth_form: growth_form === 'null' ? null : growth_form,
```
julia-rom, line 75: `growth_form: growth_form ?? null`, could work too

julia-rom (commented):
tested, seems to work. left some feedback about clarifying some var names, avoiding mutating data directly, and perhaps memoizing some of the sorting

src/components/pages/ImageClassification/ImageAnnotationModal/ImageAnnotationPopup/ExistingRows.js
julia-rom: ah ok good to know

julia-rom (approved):

PR #1037, opened 2024-09-12
M1020 new row

parmvirthind:
Relates to [Ticket 1020](https://trello.com/c/TOc9iswp/1020-edit-point-popup-new-row-modal-functionality)

**Changes:**
- Updated re-usable Modal component so use ternary instead of `&&`.
  - Using `&&` can result in a "0" being rendered to the DOM
- Updated selected row logic in `ExistingRows` component to use a `useEffect` instead of setting state in `useState` declaration. This allows the value to update when a new row is added (as seen in `NewRow`). 
- Updated `NewRow` component to open another modal when "Select New Attribute" button is clicked.
  - User can select a benthic attribute and growth form in this modal, and click "Add New Row"
  - "Add New Row" button is disabled if user has not selected a benthic attribute (growth form can be empty)
  - Clicking "Add New Row" can do 2 things:
    1. If the values the user has entered matches a classifier guess of that point, then confirm that classifier guess and move to the front of the annotations array for that point
    2. If the values the user has entered **does not** match a classifier guess, then add a new annotation, confirm the new annotation and move to front of annotations array for that point
    - Note: In both cases, any user-defined annotations for that point are removed.

### DIFF
--- a/src/components/pages/ImageClassification/ImageAnnotationModal/ImageAnnotationPopup/NewRow.js
+++ b/src/components/pages/ImageClassification/ImageAnnotationModal/ImageAnnotationPopup/NewRow.js
@@ -15,6 +15,15 @@ import Modal from '../../../../generic/Modal/Modal'
 import InputAutocomplete from '../../../../generic/InputAutocomplete'
 import { NewRowContainer, NewRowFooterContainer, NewRowLabel } from '../ImageAnnotationModal.styles'
 
+// TODO: Place this in a shared folder since used twice?
+const isAClassifierGuessOfSelectedPoint = (annotations, benthic_attribute, growth_form) =>
+  annotations.some(
+    (annotation) =>
+      annotation.is_machine_created &&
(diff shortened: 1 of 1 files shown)


src/components/pages/ImageClassification/ImageAnnotationModal/ImageAnnotationPopup/NewRow.js
```diff
@@ -42,98 +41,162 @@ const NewRow = ({
...
+      if (!is_machine_created) {
+        return acc
+      }
+
+      // Move the matching classifier guess to front of array and confirm. Unconfirm the rest.
+      if (
+        benthic_attribute === selectedBenthicAttr &&
+        growth_form === (selectedGrowthForm || null)
+      ) {
+        acc.unshift({ ...currentAnnotation, is_confirmed: true })
```
julia-rom, line 71: i think you could do a sort here instead of mutating acc

julia-rom (approved):
code looks good, left a suggestion

PR #1039, opened 2024-09-12
M1019 remove unclassify logic

parmvirthind:
Relates to [Ticket 1019](https://trello.com/c/ZWB2K5zc/1019-image-annotation-modal-remove-unclassify-button-and-logic)

**Changes**
- Removed logic relating to `is_unclassified`
- Now a point will only be unclassified if it has no annotations when returned from the API.


julia-rom (approved):
nice, this is simpler

PR #1041, opened 2024-09-16
M1017 default radio value

parmvirthind:
Relates to [Ticket 1017](https://trello.com/c/fqiXLqC4/1017-edit-point-popup-default-value-for-radio-buttons)

**Changes:**
- Removed all state relating to `selectedRadioOption`
  - Instead of having a specific state just to manage this, we instead derive this value based on the user's selection.
- Added `isAnnotationSelected` function to determine whether a classifier guess is the selected annotation for that point
- Added `_updateSelectedRowOnPointAnnotationChange` effect in `ExistingRows` component that will update the selected row anytime the selected annotation is updated. 
- Added a default disabled value of "Choose..." for `ExistingRow` dropdown

**To test:**
- Open the image annotation modal for a uploaded and processed image
- Select a point to open the Edit Point Popover.
- Ensure that the correct radio button is selected by default.

### DIFF
--- a/src/components/pages/ImageClassification/ImageAnnotationModal/ImageAnnotationPopup/ExistingRows.js
+++ b/src/components/pages/ImageClassification/ImageAnnotationModal/ImageAnnotationPopup/ExistingRows.js
@@ -47,18 +47,18 @@ const ExistingRows = ({
     return acc
   }, [])
 
-  const _updateSelectedRow = useEffect(() => {
+  const _updateSelectedRowOnPointAnnotationChange = useEffect(() => {
     const { benthic_attribute, growth_form } = selectedPoint.annotations[0]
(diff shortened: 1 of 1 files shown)


src/components/pages/ImageClassification/ImageAnnotationModal/ImageAnnotationPopup/ClassifierGuesses.js
```diff
@@ -17,12 +17,19 @@ const confirmFirstAnnotationAndUnconfirmRest = (annotation, i) => {
   annotation.is_confirmed = i === 0
 }
 
+const isAnnotationSelected = (annotation, selectedPoint) => {
```
julia-rom, line 20: nice!

julia-rom (approved):
tested, seems to work. code looks good

PR #1045, opened 2024-09-18
M1018 highlight rows

parmvirthind:
relates to [ticket 1018](https://trello.com/c/9DyW9xyl/1018-image-annotation-modal-highlight-row-and-other-points-of-same-bagr-when-hovering-on-point)

**changes**:
- when a user hovers over a row in the table it receives a blue border and all points in that row will be highlighted on the map
- when a user clicks a row in the table, it will "pin" that row as selected. it will display a blue border and highlight all points in that row. when user hovers away, the row and points stay highlighted.
- a user can un-select or un-pin a row by clicking the highlighted row.
- hover state only applies if a row is not selected/pinned.
- when a point is clicked on the map it receives a yellow, thicker border. once you click away from the point, the point goes back to its original styling.
- removed `selectedPoints` state as no longer needed.
- updated `highlightedPoints` state to be `highlightedAttributeId`. It is easier to store the ba_gr id of the  points and row that should be highlighted than storing an array of points.

**to test:**
- open image annotation modal
- hover over rows in the table and notice that it receives a blue border and that points on map are highlighting as well.
- ensure that clicking a row will lock in that selected state and persist the blue border on the row and points.
- ensure that clicking a highlighted row will remove that style once hovered away from.

### DIFF
--- a/src/components/pages/ImageClassification/ImageAnnotationModal/ImageAnnotationModalTable.js
+++ b/src/components/pages/ImageClassification/ImageAnnotationModal/ImageAnnotationModalTable.js
@@ -40,20 +40,10 @@ const ImageAnnotationModalTable = ({
     }
   }
 
-  const handleRowHover = (rowKey) => {
-    if (selectedRowKey !== undefined) {
-      return
+  const handleRowHoverOrLeave = (rowKey) => {
+    if (selectedRowKey === undefined) {
+      setHighlightedAttributeId(rowKey)
     }
-
-    setHighlightedAttributeId(rowKey)
-  }
-
-  const handleRowLeave = () => {
-    if (selectedRowKey !== undefined) {
-      return
-    }
-
-    setHighlightedAttributeId('')
   }
 
   const handleRowConfirm = (e, rowData) => {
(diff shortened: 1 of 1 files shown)


src/components/pages/ImageClassification/ImageAnnotationModal/ImageAnnotationModalTable.js
```diff
@@ -29,38 +28,49 @@ const ImageAnnotationModalTable = ({
...
-      setSelectedRowIndex(index)
-      setSelectedPoints(rowData)
+      setSelectedRowKey(rowKey)
+      setHighlightedAttributeId(rowKey)
     }
   }
 
+  const handleRowHover = (rowKey) => {
+    if (selectedRowKey !== undefined) {
+      return
```
julia-rom, line 45: could be one func:

```
 const handleRowHoverOrLeave = (rowKey = '') => {
  if (selectedRowKey === undefined) {
    setHighlightedAttributeId(rowKey)
  }
}
```

julia-rom (approved):
works well, way easier to navigate with the highlighted blue row. left a suggestion.

PR #1046, opened 2024-09-19
M1035 & M1016: Add ba_gr id and sort attributes alphabetically

parmvirthind:
Relates to Tickets [1035](https://trello.com/c/qismIzPG/1035-investigate-adding-a-bagr-property-on-points-when-fetched), [1016](https://trello.com/c/fuHiUDmD/1016-image-annotation-modal-sort-pseudo-observations-alphabetically)

**Changes:**
- Added `ba_gr` and `ba_gr_label` properties to data **after** fetched, and **before** setting to react state
  - Throughout the application we were having to do a lot of checks against `benthic_attribute` and `growth_form` separately. By combining the two into one attribute, we simplify a lot of the checks throughout the Image Annotation Modal.
  - Adding the `ba_gr_label` makes it so that we are not constantly retrieving the label for either benthic attribute or growth form via `getBenthicAttributeLabel` or `getGrowthFormLabel`. Instead we do this once at the start when data is initially fetched. 
  - Additionally in `ImageAnnotationModalTable` we can group the rows by the `ba_gr_label` attribute. We can then sort on these keys (which covers ticket 1016). Previously we were grouping by `ba_gr` which is UUID that could not be sorted.
- Removed fetching of `benthicAttributes` and `growthForms` within the `ImageAnnotationModal`. Instead pass as props from the Observation table as it is being fetched there anyways. 
  - The `ImageAnnotationModal` is updated to not open unless both `benthicAttributes` and `growthForms` have been fetched.
- Updated user annotations to include `ba_gr` and `ba_gr_label`
- Updated the dropdown options in `ExistingRows` to be sorted alphabetically.
- Updated how Benthic Attribute Growth Form label is displayed.
  - As per Kim, the correct way to display these is with a space in between and growth form a lowercase (ex. Montipora branching)

### DIFF
--- a/src/components/pages/ImageClassification/ImageAnnotationModal/ImageAnnotationModalTable.js
+++ b/src/components/pages/ImageClassification/ImageAnnotationModal/ImageAnnotationModalTable.js
@@ -62,7 +62,7 @@ const ImageAnnotationModalTable = ({ points, setDataToReview, setHighlightedAttr
           <Th colSpan={2} align="right">
             Count
           </Th>
-          <Th>Attribute / Growth Form</Th>
+          <Th>Attribute growth form</Th>
           <Th>Status</Th>
         </Tr>
       </thead>
--- a/src/components/pages/ImageClassification/ImageAnnotationModal/ImageAnnotationPopup/ExistingRows.js
+++ b/src/components/pages/ImageClassification/ImageAnnotationModal/ImageAnnotationPopup/ExistingRows.js
@@ -77,7 +77,7 @@ const ExistingRows = ({ selectedPoint, dataToReview, setDataToReview }) => {
   return (
     <>
       <Tr>
-        <Th colSpan={4}>Attribute / Growth Form</Th>
+        <Th colSpan={4}>Attribute growth form</Th>
       </Tr>
       <Tr>
         <PopupTdForRadio>


src/components/pages/ImageClassification/ImageAnnotationModal/ImageAnnotationModal.js
```diff
@@ -26,35 +26,48 @@ import { ButtonPrimary, ButtonSecondary } from '../../../generic/buttons'
...
+      : getBenthicAttributeLabel(benthic_attribute)
+
   const _fetchImageAnnotations = useEffect(() => {
     if (databaseSwitchboardInstance && projectId) {
       databaseSwitchboardInstance
         .getAnnotationsForImage(projectId, imageId)
         .then((data) => {
-          data.points.map((point) => point.annotations.sort(prioritizeConfirmedAnnotations))
-          setDataToReview(data)
+          const formattedPoints = data.points.map((point) => {
```
julia-rom, line 55: hmm we both have ways of formatting (in my upcoming table updates pr). might be good to go over it together

src/components/pages/ImageClassification/ImageAnnotationModal/ImageAnnotationPopup/NewRow.js
```diff
@@ -22,13 +22,8 @@ import {
 import { createPortal } from 'react-dom'
 
 // TODO: Place this in a shared folder since used twice?
```
julia-rom, line 24: might be good to go over this together, i think we're both repeating some stuff

julia-rom (commented):
the code looks fine - but there is now some overlap between us. particularly in reformatting annotation data

PR #1050, opened 2024-09-24
M912 delete image

parmvirthind:
Relates to **Ticket 912**

**changes:**
- call api to delete image
- update list of images after successful delete

**to test:**
- click delete button and make sure that image gets deleted from API, and removed from list.

### DIFF
--- a/src/components/pages/ImageClassification/ImageClassificationTable/ImageClassificationObservationTable.js
+++ b/src/components/pages/ImageClassification/ImageClassificationTable/ImageClassificationObservationTable.js
@@ -68,7 +68,7 @@ const statusLabels = {
 
 const ImageClassificationObservationTable = ({ uploadedFiles, setUploadedFiles }) => {
   const [imageId, setImageId] = useState()
-  const [images, setImages] = useState(uploadedFiles)
+  const [images, setImages] = useState([])
   const [polling, setPolling] = useState(false)
   const { databaseSwitchboardInstance } = useDatabaseSwitchboardInstance()
   const { projectId, recordId } = useParams()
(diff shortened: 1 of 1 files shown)


src/components/pages/ImageClassification/ImageClassificationTable/ImageClassificationObservationTable.js
```diff
@@ -353,7 +367,7 @@ const ImageClassificationObservationTable = ({ uploadedFiles, handleRemoveFile }
                         <StyledTd rowSpan={numSubRows}>
                           <ButtonCaution
                             type="button"
-                            onClick={() => handleRemoveFile(file)}
+                            onClick={() => handleRemoveImage(file)}
```
julia-rom, line 370: can we disable the button while it's being deleted?

julia-rom:
when i click the delete button, it looks like it's triggering a bunch of requests and not being removed from the table

src/components/pages/ImageClassification/ImageClassificationTable/ImageClassificationObservationTable.js
```diff
@@ -85,6 +86,19 @@ const ImageClassificationObservationTable = ({ uploadedFiles, handleRemoveFile }
     }
   }
 
+  const handleRemoveImage = (file) => {
+    databaseSwitchboardInstance
+      .deleteImage(projectId, file.id)
+      .then(() => {
+        const updatedImages = images.filter((f) => f.id !== file.id)
```
julia-rom, line 93: this seems to at least partially fix some of the glitchiness . Instead of filtering from current state, creating a new reference

`setImages((prevImages) => prevImages.filter((f) => f.id !== file.id))`

julia-rom: current bug:

1. create a new bpq record
2. upload images
3. delete image
4. many requests fire off and table glitches

if i refresh the page, and try and delete again, it works as it should

julia-rom: yes that makes sense. ok getting closer. Last thing I noticed is if I have a few images, and delete them, the last remaining image sticks around until i refresh the page.

julia-rom:
works great - approved

PR #1062, opened 2024-10-09
M1087 ic modal sizing

parmvirthind:
Relates to [Ticket 1087](https://trello.com/c/H08v64iG/1087-annotation-modal-sizing?filter=label:classification,member:parmvirthind1)

**Changes**
- Update Table in Annotation Modal to scroll vertically if list length exceeds `max-height` of `ModalContent`
- Update how the `imageScale` is calculated. 
  - Now based on both image width and height, as well as the window width and height. (prev only took into account image height)
- Added event listener for `window` resize in a `useEffect`. In this listener we re-calculate the `imageScale`
- Removed `ImageAnnotationMapContainer` styled-component. Use `style` instead of passing props
  - Was getting warning in console about too many classes being added, as the class would change anytime the window was resized.  
- Fixed glitchiness with the map "Reset" button. If window was resized and then set back to normal, clicking the button would cause the map to jump around.

**To Test**
- Open the annotation modal, and re-size your browser by both width and height and ensure the image scales.


julia-rom (approved):
map/image resizes nicely + good use of commenting !

PR #1065, opened 2024-10-10
[M897] disable ic offline

julia-rom:
[trello card](https://trello.com/c/ndfhZ4gp/897-disable-image-classification-offline)

to test:
1. create a BPQ record
2. go offline
3. see sample unit selector switch to offline text
4. go back online
5. select image classification
6. go offline again
7. see table switch to offline text

### DIFF
--- a/src/components/pages/ImageClassification/SampleUnitInputSelector/SampleUnitInputSelector.js
+++ b/src/components/pages/ImageClassification/SampleUnitInputSelector/SampleUnitInputSelector.js
@@ -1,32 +1,18 @@
 import React from 'react'
-import { ButtonPrimary, ButtonSecondary } from '../../generic/buttons'
-import { H3 } from '../../generic/text'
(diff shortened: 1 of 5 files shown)


alanjleonard (changes requested):
This should only be for image classification, not the manual entry of BPQ.
If there's a BPQ record that uses manual input that already exists, it should work offline.  
When the user creates a new BPQ while offline, just disable the button. See trello ticket for mockups.
 https://trello.com/c/ndfhZ4gp/897-disable-image-classification-offline

julia-rom:
> This should only be for image classification, not the manual entry of BPQ. If there's a BPQ record that uses manual input that already exists, it should work offline. When the user creates a new BPQ while offline, just disable the button. See trello ticket for mockups. https://trello.com/c/ndfhZ4gp/897-disable-image-classification-offline

missed something before - should display correctly now

alanjleonard (approved):

PR #1070, opened 2024-10-21
M1091 ic status indicator

parmvirthind:
[Trello Card](https://trello.com/c/gPjXqVxJ/1091-add-the-status-indicator-colour-in-the-first-column-of-bpq-observation-table)

**To test**
- Open an BPQ in IC mode. 
- Ensure that annotations that have all confirmed points displays a green indicator.
- Ensure that annotations that have at least one unconfirmed point displays a purple? indicator. 
- Ensure that images that are still being processed display a grey indicator.
- Note that [IC validation is not yet handled on the front-end](https://trello.com/c/lSlVf7Hf/1076-handle-pqt-validation-on-frontend) so testing for that aspect cannot be done, but code is in place to display the respective validation colour.


alanjleonard (approved):
Looks good. As long as the colours get updated to the colours in this PR. https://github.com/data-mermaid/mermaid-webapp/pull/1067

julia-rom (approved):
looks good

PR #1073, opened 2024-10-29
[M1076] handle validate

julia-rom:
[trello card](https://trello.com/c/lSlVf7Hf/1076-handle-pqt-validation-on-frontend)

**summary**

-  added validations column to image classification
- removed unknown/unclassified column (now represented as a subrow for an image)
-  validations belong to each specific row
-  unconfirmed data points correspond to each row
-  warning/error colors  appear for each row depending on validation status
- if unknown/unclassified points for an image exist, an additional subrow will get added for that specific image, which will display how many unclassified points there are
- hide observation validation row error if unconfirmed count = 0

**to test**

- create a BPQ record
- select image classification
- upload some images
- click validate
- see validations column appear with corresponding messages
- open annotation modal
- confirm observation
- save
- check observation table to ensure validations that previously had unconfirmed count error, and are now at 0, no longer have errors

**note that validations do NOT currently get refreshed when making changes in the annotation modal and then going back to the table, you will need to click validate again**

if unknown points exist, it looks like this @alanjleonard 

Screenshot 2024-11-06 at 5 31 57 PM

### DIFF
--- a/src/components/pages/ImageClassification/ImageClassificationTable/ImageClassificationObservationTable.js
+++ b/src/components/pages/ImageClassification/ImageClassificationTable/ImageClassificationObservationTable.js
@@ -473,21 +473,23 @@ const ImageClassificationObservationTable = ({
                             <>
                               {areValidationsShowing ? (
                                 <StyledTd>
-                                  <ObservationValidationInfo
-                                    hasObservationErrorValidation={hasObservationErrorValidation}
-                                    hasObservationIgnoredValidation={
-                                      hasObservationIgnoredValidation
-                                    }
-                                    hasObservationWarningValidation={
-                                      hasObservationWarningValidation
-                                    }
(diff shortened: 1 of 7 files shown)


src/components/pages/ImageClassification/ImageClassificationTable/ImageClassificationObservationTable.js
```diff
@@ -316,20 +359,35 @@ const ImageClassificationObservationTable = ({ uploadedFiles, setUploadedFiles }
 
   let rowIndex = 1
 
+  const buildObservationId = (imageId, benthicAttributeId = null, growthFormId = null) => {
+    let observationId = imageId
+
+    if (benthicAttributeId) {
```
julia-rom, line 365: this reflects how `observation_id` is constructed in the api

PR #1085, opened 2024-11-17
[M1134] handle excessive precision validation

saanobhaai:
## Summary by CodeRabbit

- **New Features**
	- Introduced a validation message for excessive precision in depth measurements to enhance data integrity.
- **Bug Fixes**
	- Improved validation logic to ensure depth values comply with specified precision limits.

### DIFF
--- a/src/language.js
+++ b/src/language.js
@@ -1333,6 +1333,8 @@ const getValidationMessage = (validation, projectId = '') => {
     invalid_benthic_transect: () =>
       'One or more invalid fields: site, management, sample date, transect number, width, depth',
     invalid_depth: () => `Depth invalid or not greater than ${context?.depth_range[0]} m`,
+    excessive_precision: () =>
+      `Depth precision to right of decimal point greater than ${context?.decimal_places} digit(s)`,
     invalid_fish_count: () => 'Fish count must be a non-negative integer',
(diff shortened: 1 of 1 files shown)


saanobhaai:
[capture depth decimal places in validator](https://trello.com/c/Rv0g8DCw/1134-capture-depth-decimal-places-in-validator)

src/language.js
```diff
@@ -1333,6 +1333,8 @@ const getValidationMessage = (validation, projectId = '') => {
     invalid_benthic_transect: () =>
       'One or more invalid fields: site, management, sample date, transect number, width, depth',
     invalid_depth: () => `Depth invalid or not greater than ${context?.depth_range[0]} m`,
+    excessive_precision: () =>
+      `Depth precision to right of decimal point greater than ${context?.decimal_places} digit(s)`,
```
gridcell, line 1337: Does locale change number (decimals and commas) in the UI for other countries?  If so wording may need to change

saanobhaai: No atm we're not handling locale properly. There's a card to handle currencies with US-style commas/periods in the near term; hope is to come back to locales (and char sets, and translation) in a more thorough way next year. There are a lot of other things like this to consider.

PR #1086, opened 2024-11-19
[M1155] Observation table deletion bug

julia-rom:
[trello card](https://trello.com/c/yNKDEu2E/1155-ic-observation-table-glitches-when-deleting-an-image-while-others-are-processing-queued)

bug fix for deletion bug. Previously, if you tried to delete images while others were processing or queued, the table would sometimes start to glitch, and many network requests would be fired off. This was caused by `uploadedFiles` and `images ` becoming out of sync due to a race condition. To fix, I removed `uploadFiles` var in the image classification table and replaced with 'isUploading'. I also simplified the `_startPollingOnUpload useEffect` 

to test:
1. upload images
2. delete some images that are done processing while others are uploading or queued
3. ensure table doesn't start glitching, or files don't reappear as described in the linked ticket

### DIFF
--- a/src/components/pages/ImageClassification/ImageClassificationTable/ImageClassificationObservationTable.js
+++ b/src/components/pages/ImageClassification/ImageClassificationTable/ImageClassificationObservationTable.js
@@ -92,12 +92,11 @@ const statusLabels = {
 }
 
 const ImageClassificationObservationTable = ({
-  uploadedFiles,
-  setUploadedFiles,
   collectRecord = undefined,
   areValidationsShowing,
   ignoreObservationValidations,
   resetObservationValidations,
+  isUploading,
 }) => {
   const [imageId, setImageId] = useState()
   const [images, setImages] = useState([])
@@ -158,9 +157,6 @@ const ImageClassificationObservationTable = ({
       .then(() => {
         const updatedImages = images.filter((f) => f.id !== file.id)
         setImages(updatedImages)
-
-        const updatedUploadedFiles = uploadedFiles.filter((f) => f.id !== file.id)
-        setUploadedFiles(updatedUploadedFiles)
 
         toast.warn('File removed')
       })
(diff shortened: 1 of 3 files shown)


src/components/pages/ImageClassification/ImageClassificationTable/ImageClassificationObservationTable.js
```diff
@@ -159,9 +158,6 @@ const ImageClassificationObservationTable = ({
         const updatedImages = images.filter((f) => f.id !== file.id)
         setImages(updatedImages)
 
-        const updatedUploadedFiles = uploadedFiles.filter((f) => f.id !== file.id)
```
julia-rom, line 162: we only care about updating images, no need to also keep track of `uploadedFiles`

src/components/pages/ImageClassification/ImageClassificationTable/ImageClassificationObservationTable.js
```diff
@@ -435,23 +412,23 @@ const ImageClassificationObservationTable = ({
                         <TdWithHoverText
                           data-tooltip={file.original_image_name}
                           onClick={() => handleImageClick(file)}
-                          cursor={file.classification_status.status === 3 ? 'pointer' : 'default'}
+                          cursor={file.classification_status?.status === 3 ? 'pointer' : 'default'}
```
mmnoo, line 415: I know this isnt in the scope of this diff which is just adding an `?.`, but a dev cant tell what `3` signifies here. I like to use enums or objects to make things like this more self documenting. 

eg:

```
const CLASSIFICATION_STATUSES { 
	BUMPY: 3,
	SMOOTH: 4,
	STICKY: 5,
}

...

cursor={file.classification_status?.status === CLASSIFICATION_STATUSES.sticky ? 'pointer' : 'default'}
```

mmnoo (commented):
Looks good. Some code intuitiveness comments.

src/components/pages/ImageClassification/ImageClassificationTable/ImageClassificationObservationTable.js
```diff
@@ -318,7 +314,7 @@ const ImageClassificationObservationTable = ({
         setImages(response.results)
 
         const allProcessed = response.results.every((file) =>
```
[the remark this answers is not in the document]
julia-rom: good point, renamed to `areAllImagesProcessed`

PR #1116, opened 2025-01-17
M12235 IC reset zoom button and zoom-to-point button

mmnoo:
[Ticket](https://trello.com/c/qm74tcRG/1225-image-classification-add-zoom-in-and-out-buttons-to-point-popover)
Description:
- added zoom in and out buttons
- removed semantic error in that table elements must be within a table (I had moved the select out of of the table since only the guesses can pass as appropriate to be placed in a table

Steps to test:
- open a BPQ record with image classification
- click the IC review button
- click a point in the image
- click the zoom out button => it should zoom out to the initial extent (same as reset button)
- click the zoom in button => it should zoom to the point similar to when a user clicks the point

Screenshot 2025-01-16 at 6 24 25 PM

### DIFF
--- a/src/components/generic/Table/table.js
+++ b/src/components/generic/Table/table.js
@@ -79,23 +79,23 @@ const getHeaderSortAfter = (
     }
   `
 }
+export const thStyles = (props) => css`
+  text-align: ${props.align || 'left'};
+  padding: ${theme.spacing.medium};
+  background: ${theme.color.white};
+  vertical-align: top;
+  &::after {
+    content: ${props.isSortingEnabled ? ' \u25b2' : ''};
+    font-size: small;
+    white-space: nowrap;
+  }
+  > span {
+    ${getHeaderSortAfter(props.isMultiSortColumn, props.sortedIndex, props.isSortedDescending)}
+  }
+  font-weight: 700;
+`
 
-export const Th = styled.th(
-  (props) => css`
-    text-align: ${props.align || 'left'};
-    padding: ${theme.spacing.medium};
-    background: ${theme.color.white};
-    vertical-align: top;
-    &::after {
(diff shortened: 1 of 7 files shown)


mmnoo:
[Image classification: add zoom in and out buttons to point popover](https://trello.com/c/qm74tcRG/1225-image-classification-add-zoom-in-and-out-buttons-to-point-popover)

alanjleonard (changes requested):
- If the screen/image is small enough that the image doesn't fill the container, both buttons seems to zoom out when the screen is smaller. See screen shots. Note the screen size at the top of the screnshots.
image
image

- Buttons need hover states (addressed in Trello)
- Is removing the auto-zoom when you click a point in another ticket?

mmnoo:
@alanjleonard I made a hover ticket. If you would like auto zoom disabled pls make a ticket if one doesnt already exist.

I suspect the small screen zooming out is because we are trying to make sure the popup fits in the image too... will play with it and get back to you.

mmnoo:
> @alanjleonard I made a hover ticket. If you would like auto zoom disabled pls make a ticket if one doesnt already exist.
> 
> I suspect the small screen zooming out is because we are trying to make sure the popup fits in the image too... I will play with it and get back to you.

yup @alanjleonard , the zoom in on small screens actually zooming out is because we have some extra padding around the point patch so that the popover is likely to still be visible. In a small screen this can result in zooming out behaviour. Lets figure out what we want and the tradeoffs in another ticket. We can choose to cut off more of the popover so that the zoom in is actually zooming in on small screens (right now we are technically 'fitting the bounds' of the patch plus padding to match the point click behaviour, so thats why its not necessarily a zoom in operation).

PR #1117, opened 2025-01-17
M1238 IC popup table hover styling

mmnoo:


mmnoo:
[give icon buttons hover state and other styles to be similar to button secondary](https://trello.com/c/fJnZ5JJB/1238-give-icon-buttons-hover-state-and-other-styles-to-be-similar-to-button-secondary)

PR #1119, opened 2025-01-18
M1226 IC next unconfirmed button

mmnoo:
[Ticket](https://trello.com/c/ocCH6g2a/1226-image-classification-add-next-confirmed-point-button)

Steps to test:
- open a BPQ record with image classification
- click the IC review button
- click on a point in the image
- hover over the arrow button beside the confirm button => you should see a tooltip
- click the button, it should zoom you to the next unconfirmed (classified or unclassified) point (the zoom is weird due to constraints with us hacking an image onto a map)
- confirm all the points on the map => the next unconfirmed point button should now be disabled

### DIFF
--- a/src/components/TransferSampleUnitsModal/TransferSampleUnitsModal.js
+++ b/src/components/TransferSampleUnitsModal/TransferSampleUnitsModal.js
@@ -4,7 +4,7 @@ import styled from 'styled-components'
 
 import { ButtonPrimary, ButtonSecondary } from '../generic/buttons'
 import language from '../../language'
-import { IconArrowRight } from '../icons'
+import { IconArrowRightCircle } from '../icons'
 import { Select } from '../generic/form'
 import { Column } from '../generic/positioning'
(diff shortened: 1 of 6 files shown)


mmnoo:
@alanjleonard I made an exec decision to remove auto closing the point popup on confirm click. This way we can keep clicking the next unconfirmed point button as we confrim points. If the popup closed on confirm, we never really get a chance to use the next unconfirmed button unless we dont confirm anything.

Also the zoom is weird. I can demo what it feels like if we remove zooming to the next feature. Unfortunately due to hacks to make an image use mapbox, we need to zoom out first to calculate the position of the popup and zoom bounds. Perhaps letting a user stay zoomed with an option to use use the zoom in button is worth considering, or maybe the as-is auto zoom is good enough.

mmnoo:
[Image classification: add next confirmed point button](https://trello.com/c/ocCH6g2a/1226-image-classification-add-next-confirmed-point-button)

alanjleonard (approved):

PR #1128, opened 2025-01-27
M1244 fix polling after ic upload so that it doesnt hammer api with excessive calls

mmnoo:
Steps to test:
- go to a BPQ record with IC
- upload multiple images with network speed throttled (4g sufficient)
- look in the network tab => polling for all images for the collect record should happen no more frequently than 5 second intervals

### DIFF
--- a/src/components/pages/ImageClassification/ImageClassificationTable/ImageClassificationContainer.js
+++ b/src/components/pages/ImageClassification/ImageClassificationTable/ImageClassificationContainer.js
@@ -4,15 +4,64 @@ import ImageUploadModal from '../ImageUploadModal/ImageUploadModal'
 import { ButtonPrimary } from '../../../generic/buttons'
 import { IconUpload } from '../../../icons'
 import { ButtonContainer, IconContainer } from './ImageClassificationObservationTable.styles'
+import { EXCLUDE_PARAMS_FOR_GET_ALL_IMAGES_IN_COLLECT_RECORD } from '../imageClassificationConstants'
+import { getIsImageProcessed } from '../getIsImageProcessed'
+import { useDatabaseSwitchboardInstance } from '../../../../App/mermaidData/databaseSwitchboard/DatabaseSwitchboardContext'
+import { useHttpResponseErrorHandler } from '../../../../App/HttpResponseErrorHandlerContext'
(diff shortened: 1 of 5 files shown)


src/components/pages/ImageClassification/ImageClassificationTable/ImageClassificationContainer.js
```diff
@@ -4,23 +4,74 @@ import ImageUploadModal from '../ImageUploadModal/ImageUploadModal'
...
+          projectId,
+          recordId,
+          EXCLUDE_PARAMS_FOR_GET_ALL_IMAGES_IN_COLLECT_RECORD,
+        )
+
+        setImages(response.results)
+
+        const areAllImagesProcessed =
+          response.results.length &&
+          response.results.every((file) => getIsImageProcessed(file.classification_status?.status))
```
mmnoo, line 45: This every passed if using an empty array! So we needed to also check for array length if we want to be able to use this function recursively (re poll the collect records). Crazy JS!

mmnoo: (this is mostly copied code with some minor renaming from the other file, sans being triggered by an effect. Its now triggered by user upload directly)

src/components/pages/ImageClassification/ImageClassificationTable/ImageClassificationObservationTable.js
```diff
@@ -319,53 +315,6 @@ const ImageClassificationObservationTable = ({
...
-    if (polling) {
-      startPolling()
-    }
-
-    return () => {
-      if (intervalId) {
-        clearTimeout(intervalId)
-      }
-    }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
```
mmnoo, line 366: this effect was overreactive causing excessive polling, even with ignored deps (This is a lint rule I never ignore)

mmnoo:
[Look into network traffic and errors from IC upload](https://trello.com/c/4vn5vvEq/1244-look-into-network-traffic-and-errors-from-ic-upload)

mmnoo:
Merging this for descendent branch manageability. Please still review!

PR #1129, opened 2025-01-28
Image classification/m1256 missing error handler

mmnoo:
Just being consistent and through :) 

Steps to test: 
- its unlikely this error will trigger due to network issues since the values also live in indexeddb. (network failure would more likely happen on initial app load or sync pulls than the code in this diff)


mmnoo:
[use error handler for network call missed in other ticket](https://trello.com/c/JYsBIVXP/1256-use-error-handler-for-network-call-missed-in-other-ticket)

src/components/pages/ImageClassification/useSelectNewAttribute.js
```diff
@@ -1,6 +1,7 @@
 import { useState } from 'react'
 import { getBenthicOptions } from '../../../library/getOptions'
 import { toast } from 'react-toastify'
+import { useHttpResponseErrorHandler } from '../../../App/HttpResponseErrorHandlerContext'
```
JBurkinshaw, line 4: Nit pick. Keep the 3rd party imports separate from imports from the application

src/components/pages/ImageClassification/useSelectNewAttribute.js
```diff
@@ -30,8 +32,14 @@ export const useSelectNewAttribute = ({
         setGrowthFormSelectOptions(choices.growthforms.data)
         setShouldDisplayModal(true)
       })
-      .catch(() => {
-        toast.error('Failed to retrieve benthic attributes and growth forms')
+      .catch((error) => {
+        handleHttpResponseError({
+          error,
+          callback: () => {
+            toast.error('Failed to retrieve benthic attributes and growth forms.', error.message)
```
JBurkinshaw, line 39: Does this need to be in the language file?

src/components/pages/ImageClassification/useSelectNewAttribute.js
mmnoo: Agree stylistically. :)

PR #1136, opened 2025-01-31
M1169 add collect record status column to records table, as well as border colour for each status

mmnoo:
[Ticket](https://trello.com/c/YslpaBlf/1169-show-cr-validation-status-in-collecting)

Steps to test:
- go to a project's collect record page
- on the right side of the table, see the new 'status' column
- on the left side of the table notice the new ~~border-right~~ border-left colour that corresponds with status
- click on the new status column => it should be sortable

Screenshot 2025-01-31 at 4 07 09 PM

## Summary by CodeRabbit

- **New Features**
  - Added a status column to the Collect page table
  - Implemented visual differentiation for record statuses
  - Introduced user-friendly status labels for records

- **Improvements**
  - Enhanced table row rendering to display record status
  - Added color-coded status indicators for better data visualization

### DIFF
--- a/src/components/pages/Collect/Collect.js
+++ b/src/components/pages/Collect/Collect.js
@@ -45,6 +45,8 @@ import {
 } from '../../../App/mermaidData/recordProtocolHelpers'
 import { getIsUserReadOnlyForProject } from '../../../App/currentUserProfileHelpers'
 import { PAGE_SIZE_DEFAULT } from '../../../library/constants/constants'
+import { RECORD_STATUS_LABELS } from './collectConstants'
+import { TrCollectRecordStatus } from './Collect.styles'
 
 const Collect = () => {
   const [collectRecordsForUiDisplay, setCollectRecordsForUiDisplay] = useState([])
@@ -138,13 +140,18 @@ const Collect = () => {
         accessor: 'observers',
         sortType: reactTableNaturalSort,
       },
+      {
+        Header: 'Status',
+        accessor: 'status',
+        sortType: reactTableNaturalSort,
+      },
     ],
     [],
   )
 
   const tableCellData = useMemo(
     () =>
-      collectRecordsForUiDisplay.map(({ id, data, uiLabels }) => {
+      collectRecordsForUiDisplay.map(({ id, data, uiLabels, validations }) => {
         const isQuadratSampleUnit = getIsQuadratSampleUnit(data.protocol)
 
         return {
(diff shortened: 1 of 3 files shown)


src/components/pages/Collect/Collect.js
```diff
@@ -45,6 +45,8 @@ import {
 } from '../../../App/mermaidData/recordProtocolHelpers'
 import { getIsUserReadOnlyForProject } from '../../../App/currentUserProfileHelpers'
 import { PAGE_SIZE_DEFAULT } from '../../../library/constants/constants'
+import { RECORD_STATUS_LABELS } from './collectConstants'
```
JBurkinshaw, line 48: Are you happy with this approach where constants are kept in different places?

src/components/pages/Collect/Collect.styles.js
```diff
@@ -0,0 +1,15 @@
...
+import { Tr } from '../../generic/Table/table'
+import styled from 'styled-components'
+import theme from '../../../theme'
+
+const RECORD_STATUS_COLORS = {
+  [RECORD_STATUS_LABELS.error]: theme.color.cautionColor,
+  [RECORD_STATUS_LABELS.ok]: '#298217',
+  [RECORD_STATUS_LABELS.stale]: theme.color.primaryColor,
+  [RECORD_STATUS_LABELS.warning]: theme.color.warningColor,
+}
```
JBurkinshaw, line 11: I think I like this approach but it took me a minute to understand it.

Also, why is colour a hex colour, whereas the others are from the theme?

JBurkinshaw (commented):
In the description do you mean In the description, do you mean `border-left` instead of `border-right`?

src/components/pages/Collect/Collect.js
mmnoo: Im really either or about it. Was just acting out of habit/auto pilot.  I guess the fact that they are removed makes it harder to read/understand and these ones are unlikely to be reused.

src/components/pages/Collect/Collect.styles.js
mmnoo: Al was specific that he wanted that colour. Im not sure yet how it ties into the theme since its used nowhere else (I search)

I think its good to put certain values into constants - like map layer names for example where you are using them a lot to get and set. It takes away most of the typo bugs, and then you get autocomplete. Also in this case, if the api changes the ‘ok’ status to something else, I can just change it in one place (theoretically). 

Can you think of a way that is easier to understand that still uses constants? Are the value of constants not worth it in you mind?

alanjleonard (changes requested):
Using it now, it should be "Errors" and "Warnings", plural in the status column. Otherwise great!

mmnoo:
> In the description do you mean In the description, do you mean `border-left` instead of `border-right`?

Yes!

mmnoo:
[show CR validation status in Collecting](https://trello.com/c/YslpaBlf/1169-show-cr-validation-status-in-collecting)

PR #1142, opened 2025-02-11
M1056 audit getting all paginated notifications

mmnoo:
[Ticket](https://trello.com/c/Znwe6wcB/1056-audit-and-handle-pagination-better)

Get all paginated notifications

Steps to test
- add a user to a project, then delete them. This ensures there will be a couple of notifications to test. You should see a notification badge in the header, and notifications in the notification dropdown. 

Note: This tests that we didnt break anything, however, there arent reasonably enough notifications to test that the recursive function works, however, this was manually tested by setting results per page very low in the code. Here we are testing that we havent broken any existing functionality.


src/App/bellNotificationHelpers.js
```diff
@@ -1,6 +1,7 @@
 import axios from '../library/axiosRetry'
```
mmnoo, line 1: Most of this diff is just adjusting notifications to operate on the notifications array instead of the `response.data` object that the API returns (which contains the notifications array on its `results` property). This is so we can use a utility function previously developed to get all the API 'pages' of data without rewriting the logic for it.

src/App/mermaidData/databaseSwitchboard/ProjectsMixin.js
```diff
@@ -80,7 +80,7 @@ const ProjectsMixin = (Base) =>
         authorizationHeaders: await getAuthorizationHeaders(this._getAccessToken),
         axios,
         errorCallback: () => {
-          Promise.reject(this._notAuthenticatedAndReadyError)
+          return Promise.reject(this._notAuthenticatedAndReadyError)
```
mmnoo, line 83: As promised in last PR

mmnoo:
[audit and handle pagination better](https://trello.com/c/Znwe6wcB/1056-audit-and-handle-pagination-better)

PR #1144, opened 2025-02-12
M1278 ensure GFCR f2.5 input value gets sent to API

mmnoo:
[Ticket](https://trello.com/c/dsSOaOXv/1278-gfcr-sub-indicator-f25-not-saving-with-actual-value-entered) <= includes steps to test


mmnoo:
> tested locally and changes to f2.5 persists.
> 
> overall though, wondering if F2Form component could be re-factored to be DRYer. Don't get me wrong I'm not DRY above all else, but this seems like a candidate at a quick glance

I think I agree, although Im not going to do it in this ticket! Although, if any fields need special treatment or exceptions the DRY would be at risk of being the wrong kind. I think tedious verbosity here could also be appropriate if there are any fields that break the pattern. (I dont actually know because its the first time I touched this feature and am just learning what its all about)

PR #1150, opened 2025-02-15
M377 use dashes for empty but required values in specified tables

mmnoo:
[Ticket](https://trello.com/c/8fJnGx7i/377-make-use-of-dash-consistent)

To test check that:
- empty values in collect record table have dashes (click on 'collecting' in side nav)
- empty values in observers and transects page (click on same name in side nav to view) show dashes, but only if they have a yellow background
- no empty values in management regimes overview page (click on same name in side nav to view) have dashes. 

## Summary by CodeRabbit

- **New Features**
  - Improved table display logic to consistently handle empty or whitespace-only values by showing a designated placeholder or empty value across key pages.
  
- **Tests**
  - Introduced comprehensive tests to ensure robust detection of empty content across various data types.

### DIFF
--- a/src/library/getIsEmptyStringOrWhitespace.js
+++ b/src/library/getIsEmptyStringOrWhitespace.js
@@ -1,0 +1,7 @@
+export const getIsEmptyStringOrWhitespace = (value) => {
+  if (typeof value !== 'string') {
+    return false
+  }
+
+  return value.trim() === ''
+}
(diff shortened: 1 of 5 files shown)


src/library/getIsEmptyStringOrWhitespace.js
```diff
@@ -0,0 +1,7 @@
+export const getIsEmptyStringOrWhitespace = (value) => {
```
JBurkinshaw, line 1: What if a valid value with a different type like `number` is passed in?

JBurkinshaw (approved):

src/library/getIsEmptyStringOrWhitespace.js
mmnoo: It returns false because its not typeof = string. (verified with unit tests!)

mmnoo:
[make use of dash consistent](https://trello.com/c/8fJnGx7i/377-make-use-of-dash-consistent)

PR #1151, opened 2025-02-19
M406 fix user icon displaying image, then initials, and finally a fallback icon.

mmnoo:
[Ticket](https://trello.com/c/c8HH1dc7/406-avatar-image-from-google-missing?)

Description of changes:
- add a `useLayoutEffect` to clear any error state from attempts to render the user icon with an image (previously once there was one error, it would persist no matter what)
- change user icon logic to use initials instead of full names (to fit in the header)

I noticed the google user image is rate limited and its very easy to hit the limit, so I looked into rerenders, but found no real low hanging fruit to improve app or header rerenders (seems the header has no obviously unnecessary rerenders). 

## Summary by CodeRabbit

- **New Features**
	- Enhanced user profile display in the header to show profile images, user initials, or a default icon based on available data.
- **Style**
	- Introduced a new circular design for the user icon, featuring refined typography and balanced dimensions for a modern appearance.
- **Bug Fixes**
	- Updated expected values in tests to reflect changes in user initials for both online and offline scenarios.
	- Adjusted user name expectations in various test cases to align with updated user data format.

### DIFF
--- a/src/components/Header/Header.js
+++ b/src/components/Header/Header.js
@@ -1,9 +1,9 @@
 import { Link } from 'react-router-dom'
 import PropTypes from 'prop-types'
-import React, { useState } from 'react'
+import React, { useLayoutEffect, useState } from 'react'
 
 import {
-  AvatarWrapper,
+  UserButton,
   GlobalNav,
   HeaderButtonThatLooksLikeLink,
   StyledHeader,
@@ -21,9 +21,10 @@ import {
   BiggerIconMenu,
   LoggedInAs,
   HeaderIconWrapper,
+  UserCircle,
 } from './Header.styles'
 import { currentUserPropType } from '../../App/mermaidData/mermaidDataProptypes'
-import { IconDown, IconGlobe, IconLibraryBooks } from '../icons'
+import { IconGlobe, IconLibraryBooks } from '../icons'
 import { useBellNotifications } from '../../App/BellNotificationContext'
 import { useOnlineStatus } from '../../library/onlineStatusContext'
 import BellNotificationDropDown from '../BellNotificationDropDown/BellNotificationDropDown'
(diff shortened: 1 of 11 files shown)


mmnoo:
@al, you may want to play with this one when you are back (on develop)

src/components/Header/Header.js
```diff
@@ -103,66 +120,30 @@ const Header = ({ logout = () => {}, currentUser = undefined }) => {
...
+  const fallbackButton = (
+    <AvatarWrapper>
+      <BiggerIconUser />
+    </AvatarWrapper>
+  )
 
-    if (currentUser && currentUser.first_name) {
-      return currentUser.full_name
-    }
+  const userIconButton = profileImageButton || initialsButton || fallbackButton
```
mmnoo, line 144: Ill abstract this into a user icon component in upcoming work. Turns out similar issues happen in the users table too.

mmnoo:
[Avatar image from Google missing](https://trello.com/c/c8HH1dc7/406-avatar-image-from-google-missing)

src/components/Header/Header.js
```diff
@@ -103,66 +120,30 @@ const Header = ({ logout = () => {}, currentUser = undefined }) => {
...
-          {currentUser && currentUser.full_name} <IconDown />
-        </AvatarWrapper>
-      )
-    }
-
-    // User icon
-    return (
+  const profileImageButton = isProfileImageButtonShowing ? (
+    <AvatarWrapper>
+      <CurrentUserImg src={currentUser?.picture} alt="" onError={handleImageError} />
```
JBurkinshaw, line 125: The `alt` property should have a value which isn't an empty string

mmnoo: Great catch! Thanks!

mmnoo: I should add aria label to all the other user icons too if that wasnt previously built

PR #1173, opened 2025-03-12
M1336 speculative fix of cannot read properties of undefined (not abl…

mmnoo:
…e to reproduce bug)


mmnoo:
[Project info page doesn't load on non-recent projects.](https://trello.com/c/mtSwpUYi/1336-project-info-page-doesnt-load-on-non-recent-projects)

PR #1174, opened 2025-03-12
Image classification/m1342 make benthic attribute options viewable

mmnoo:


mmnoo:
[IC: Need to ensure select new attributes autocomplete options are viewable](https://trello.com/c/OozBNtYI/1342-ic-need-to-ensure-select-new-attributes-autocomplete-options-are-viewable)

PR #1187, opened 2025-03-18
M1369 cleanup following M1367 to make number formatting behaviour consistent and easy to understand

mmnoo:
[Ticket](https://trello.com/c/fjXT9FAs/1369-do-necessary-non-functional-code-clean-up-for-1367-merge-after-march-19)

Leave testing to QA if you dont want to test almost all of GFCR, otherwise see ticket for testing details.


src/components/pages/gfcrPages/GfcrIndicatorSetForm/GfcrIndicatorSetForm.js
```diff
@@ -22,18 +21,6 @@ const StyledForm = styled.form`
...
-  if (value.trim() === '') {
-    setInputToDefaultValue(formik, fieldName)
-  }
-  const valueAsNumber = Number(value)
-
-  if (value && isRoundTo1DP && !Number.isInteger(valueAsNumber)) {
-    formik.setFieldValue(fieldName, Number(roundToOneDecimal(valueAsNumber)))
-  }
-}
-
```
mmnoo, line 36: It was unclear what was happening where this was called (onBlur events within child components). We now call more descriptively named functions to restrict decimal places for input values or replace user inputted blanks with previously saved/default values.

 Also the functionality can be called directly from generic libraries and doesnt need to live within a parent.

src/components/pages/gfcrPages/GfcrIndicatorSetForm/subPages/F1Form.js
```diff
@@ -25,7 +26,9 @@ const F1Form = ({ formik, displayHelp, handleInputBlur, handleInputFocus }) => {
         type="number"
         unit="km²"
         {...formik.getFieldProps('f1_1')}
-        onBlur={(event) => handleInputBlur(formik, event, 'f1_1')}
+        onBlur={(event) =>
+          resetEmptyFormikFieldToInitialValue({ formik, event, fieldName: 'f1_1' })
```
mmnoo, line 30: now there is less of a mystery what is going on.

src/components/pages/gfcrPages/GfcrIndicatorSetForm/subPages/F4Form.js
```diff
@@ -217,12 +235,19 @@ const F4Form = ({
           type="number"
           unit="kg/ha"
           {...formik.getFieldProps('f4_3')}
-          onBlur={(event) => handleInputBlur(formik, event, 'f4_3', true)}
```
mmnoo, line 220: sometimes calling this function was done with a fourth unnamed parameter. What was hidden there is that passing true as the 4th param was reformatting the input to only have 1 decimal. I opted to use the existing decimal restriction function to stay consistent in code and behaviour, so now onBlur only deals with resetting defaults. (the new code wont allow the user to type too many decimals, the old way just overwrote whatever they input to have less decimals after they clicked away so this is more consistent in terms of UX too now)

src/library/formik/formikHandleInputTypes.js
```diff
@@ -0,0 +1,31 @@
...
+  fieldName,
+  maxNumberOfDecimals = 2,
+}) => {
+  formik.setFieldValue(
+    fieldName,
+    getNumberTruncatedToXDecimalPlaces({
+      inputValue: event.target.value,
+      maxNumberOfDecimals,
+    }),
+  )
```
mmnoo: this is actually a renamed file. The lines hilighted here were reduced from many many more and put into a reusable function (that I can unit test which is easier than manual dev testing in this case)

src/library/numbers/getValueTruncatedToXDecimalPlaces.js
```diff
@@ -0,0 +1,35 @@
+export const getValueTruncatedToXDecimalPlaces = ({ inputValue, maxNumberOfDecimals = 2 }) => {
+  // will return a number if a number is used as the input, and a string if a string is used as the input
```
mmnoo, line 2: Unfortunately we cant use the vite refactor with TS just yet! Soon though!

src/library/numbers/getValueTruncatedToXDecimalPlaces.test.js
```diff
@@ -0,0 +1,81 @@
+import { getValueTruncatedToXDecimalPlaces } from './getValueTruncatedToXDecimalPlaces'
```
mmnoo, line 1: for simple unit tests with a little supervision copilot is pretty good at the very least to save a ton of typing IMO. I havent tested it with proper component/integration tests yet.

JBurkinshaw (approved):

PR #1218, opened 2025-04-15
[M1403] change dashboard links in Collect

tienhoah:


alanjleonard (approved):
Looks good! Only thing would be make the modal wider on smaller screens so it doesn't have to scroll horizontally. Otherwise approved.

tienhoah:
> * Clicking the link to the sample event should open the single-site metric pane as well as select the site (some work, some don't)
> * Clicking the link to view all my projects did no enable the "your projects only" filter

1) I think your env for api database is different between Collect and Explore apps.
2) The current Explore app doesn't have this feature, but I will still add the flag to this Collect app first, then I create a feature branch to enable this in Explore app.

hurrendor (approved):

PR #1230, opened 2025-05-01
M1390

hurrendor:
Also contains the commit for 1430: https://trello.com/c/bna2lnfe/1430-update-the-order-of-the-legend-in-the-image-annotation-modal

## Summary by CodeRabbit

- **Style**
  - Updated the order of legend items in the annotation modal footer for improved clarity.
  - Enhanced popup content styling with a new font family and increased font size for better readability.

### DIFF
--- a/src/components/pages/ImageClassification/ImageAnnotationModal/ImageAnnotationPopup/ImageAnnotationPopup.css
+++ b/src/components/pages/ImageClassification/ImageAnnotationModal/ImageAnnotationPopup/ImageAnnotationPopup.css
@@ -1,9 +1,14 @@
 .edit-point-popup > .maplibregl-popup-content {
-  padding: 0;
-  margin-top: -5px;
+    padding: 0;
+    margin-top: -5px;
 }
 
 /* Remove Tip */
 .edit-point-popup > .maplibregl-popup-tip {
-  display: none;
+    display: none;
+}
+
+.maplibregl-popup-content {
+    font-family: 'Open Sans', 'Helvetica Neue', 'Helvetica', 'Arial', 'sans-serif';
+    font-size: 1.6rem;
 }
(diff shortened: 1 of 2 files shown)


hurrendor:
[Use the default font in the image annotation pop over.](https://trello.com/c/Ed0R7qhZ/1390-use-the-default-font-in-the-image-annotation-pop-over)

alanjleonard (approved):

src/components/pages/ImageClassification/ImageAnnotationModal/ImageAnnotationPopup/ImageAnnotationPopup.css
```diff
@@ -1,9 +1,14 @@
...
 }
 
 /* Remove Tip */
 .edit-point-popup > .maplibregl-popup-tip {
-  display: none;
+    display: none;
+}
+
+.maplibregl-popup-content {
+    font-family: 'Open Sans', 'Helvetica Neue', 'Helvetica', 'Arial', 'sans-serif';
```
mmnoo, line 12: Non-blocking: I wonder if we can set this as a css var that JS can read with something like `getPropertyValue` and styled-components `createGlobalStyle` in `globalStyles.js` will honour it as well as the css here so we arent setting it in two places. 

Out-of-scope thought: I wonder now since [styled components is going into maintenance mode](https://opencollective.com/styled-components/updates/thank-you), with future support likely to wane, what our succession plan should be.

mmnoo:
Approved (hit wrong option above)

src/components/pages/ImageClassification/ImageAnnotationModal/ImageAnnotationPopup/ImageAnnotationPopup.css
hurrendor: That's a great call-out, and with that insight I would probably avoid adding more implementations to rely on styled-components. This is definitely a hacky fix to override the MapLibre global values but I'll look into rewriting it with some of the refactoring work being done in finalizing image classification

PR #1232, opened 2025-05-02
[M1459] Zoom to the map extent using the project’s bbox or the sample event’s lat/lng

tienhoah:
## Summary by CodeRabbit

- **New Features**
  - Explore button and links now include bounding box parameters when available, providing more context-specific navigation.
- **Bug Fixes**
  - Explore buttons are now hidden when the app is offline, ensuring accurate button visibility.
  - Explore button rendering now requires a valid sample event ID, preventing invalid launches.
- **Refactor**
  - Improved URL construction for Explore links for greater flexibility and reliability.
  - Consolidated project-related state for cleaner component logic.
  - Extracted Explore link opening logic into a reusable utility function.
- **Chores**
  - The Explore feature is now enabled by default for users unless explicitly disabled.
  - Accessibility labels for Explore buttons updated to use localized text strings.
  - Minor syntax fixes in localization files.

### DIFF
--- a/src/components/ProjectName/ProjectName.jsx
+++ b/src/components/ProjectName/ProjectName.jsx
@@ -88,7 +88,7 @@ const ProjectName = () => {
       <MuiTooltip title={language.pages.gotoExplore('this project')} placement="top" arrow>
         <IconButton
           type="button"
-          aria-label="View MERMAID Explore"
+          aria-label={language.accessibilityText.viewMERMAIDExplore}
           onClick={handleExploreButtonClick}
         >
           <BiggerIconGlobe />
--- a/src/components/ProjectName/ProjectName.jsx
+++ b/src/components/ProjectName/ProjectName.jsx
@@ -1,5 +1,5 @@
-import { useParams } from 'react-router-dom'
 import React, { useState, useEffect } from 'react'
+import { useParams } from 'react-router-dom'
 import { toast } from 'react-toastify'
 import styled, { css } from 'styled-components'
 import theme from '../../theme'
@@ -10,6 +10,7 @@ import useIsMounted from '../../library/useIsMounted'
 import { useCurrentUser } from '../../App/CurrentUserContext'
 import { useExploreLaunchFeature } from '../../library/useExploreLaunchFeature'
 import { getToastArguments } from '../../library/getToastArguments'
+import { useOnlineStatus } from '../../library/onlineStatusContext'
 import { IconGlobe } from '../icons'
 import { MuiTooltip } from '../generic/MuiTooltip'
 import { IconButton } from '../generic/buttons'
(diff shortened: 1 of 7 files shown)


src/components/ProjectName/ProjectName.jsx
```diff
@@ -47,57 +48,68 @@ const BiggerIconGlobe = styled(IconGlobe)`
...
+
+    if (!isAppOnline || isTestProject) {
+      return null
+    }
+
+    const exploreButtonContent = isExploreLaunchEnabledForUser ? (
+      <MuiTooltip title={language.pages.gotoExplore('this project')} placement="top" arrow>
+        <IconButton
+          type="button"
+          aria-label="View MERMAID Explore"
```
hurrendor, line 93: Request: Can this be a language token?

src/components/ProjectName/ProjectName.jsx
```diff
@@ -47,57 +48,68 @@ const BiggerIconGlobe = styled(IconGlobe)`
...
-      </IconButton>
-    </MuiTooltip>
-  ) : (
-    <ProjectNameLink href={`${mermaidExploreLink}/?project=${projectName}`} target="_blank">
-      <IconGlobe />
-      <span>{language.pages.goToDashboard}</span>
-    </ProjectNameLink>
-  )
+  const renderExploreButton = () => {
+    const isTestProject = project?.status < 90
```
hurrendor, line 83: Question: What does 90 signify?

src/components/ProjectName/ProjectName.jsx
```diff
@@ -47,57 +48,68 @@ const BiggerIconGlobe = styled(IconGlobe)`
...
   }, [databaseSwitchboardInstance, isMounted, projectId])
 
   const handleExploreButtonClick = () => {
-    if (!projectName) {
-      toast.error(...getToastArguments(language.error.noProjectMermaidExplore))
-      return
+    const { name, bbox } = project
+    const queryParams = new URLSearchParams({ project: name })
+
+    if (bbox) {
```
hurrendor, line 75: For :73 - :76, with it being re-used in pages/Projects, can it be extracted into a helper function?

hurrendor (changes requested):

alanjleonard (approved):

src/components/ProjectName/ProjectName.jsx
tienhoah: I don't think we do it for `aria-label`

src/components/ProjectName/ProjectName.jsx
tienhoah: Codes for each status
```
90 = open
80 = test
10 = locked
```

src/components/ProjectName/ProjectName.jsx
tienhoah: added reusable `openExploreLinkWithBbox` helper function

src/components/ProjectName/ProjectName.jsx
hurrendor: Is this already somewhere in the code as documentation? Could you add a comment here to signify that if not?
Also, should this only be targeting test projects, or is it intentionally also stating locked projects as test projects?

src/components/ProjectName/ProjectName.jsx
hurrendor: We can start adding them in, they'll also need to be translated or updated. Any text that a user might see or work with should go into the translation file

tienhoah: it's specifically used for accessibility, so I think we don't do it in the first place.
May be @alanjleonard has more say in this matter, do you want us to put all aria-label text inside language file?

I’m happy to take care of it, but I suggest handling the accessibility text updates in a separate ticket where I can move everything into the language files. That way, this ticket stays focused and doesn’t become too overwhelming.

src/components/ProjectName/ProjectName.jsx
hurrendor: Nice, thanks!

src/components/ProjectName/ProjectName.jsx
hurrendor: @tienhoah One of the long term plans is having all of the hard coded text to be tokenized, including aria-labels. It would be helpful if we are adding them as we go, with less overall effort and less error-prone than having one ticket to bulk update everything. Please add this string to the language file
```suggestion
          aria-label={language.userMessage.viewExplorer}
```

```language.tsx
const userMessage = {
 viewExplorer: 'View MERMAID Explore'
} 
```

tienhoah: @hurrendor I added the separate object used for accessibility text, we could build that up more in the future.

alanjleonard: Wouldn't it be the same as the label which would be in the language file?

tienhoah: @alanjleonard  I don't see this one existed in language file.

hurrendor: @alanjleonard Are you asking under the assumption of having the same text for the tooltip as what would be on the ARIA help text?

hurrendor (approved):

PR #1245, opened 2025-05-21
M1486

hurrendor:


alanjleonard (changes requested):
- Confirming a patch should change the dotted border to a solid border
- Let's lose the dot in the hover state. I don't think it needs to be there now 
image

- In the mockup, the border for the unclassified was solid, but seeing this it seems to make more sense if it was dotted. So let's leave that as well.

alanjleonard (approved):

PR #1258, opened 2025-05-31
fix: [M1470] Benthic PIT "Use Interval Size as Interval Start" checbox

tienhoah:


alanjleonard (changes requested):
This still doesn't seem to work. If the `Use Interval Size as Interval Start` is checked, then the two values in `Interval Size` and `Interval Start` should always be in sync. So if I change Interval size while the check box is checked, then interval Start should change to the same value.

alanjleonard (approved):

hurrendor (approved):

PR #1282, opened 2025-07-03
fix: lint from merging M1522 & M1545

tienhoah:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stability of the image annotation map by ensuring consistent cleanup behavior when the map is not loaded.


src/components/pages/ImageClassification/ImageAnnotationModal/ImageAnnotationModalMap.tsx
hurrendor, line 1: When this is linted locally, do any of the following return statements also bring up a linting issue? It's a bit odd that it would only target one item

hurrendor (approved):

PR #1287, opened 2025-07-04
fix: [M1562] Interval size not updating for SUs that already exist.

tienhoah:
Trello: https://trello.com/c/6h3DHPgx/1562-interval-size-not-updating-for-sus-that-already-exist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved form behavior for interval size and synchronization checkbox in the Benthic Pit Transect form to ensure more consistent and atomic updates.

* **Chores**
  * Removed an unused ESLint comment from the image annotation modal component.


src/components/pages/collectRecordFormPages/BenthicPitForm/BenthicPitTransectInputs.jsx
```diff
@@ -198,10 +198,18 @@ const BenthicPitTransectInputs = ({
...
+    const newIntervalSizeValue = event.target.value
+
+    const newValues = {
+      ...formik.values,
+      interval_size: newIntervalSizeValue,
+    }
+
     if (isIntervalSizeAsStartChecked) {
-      formik.setFieldValue('interval_start', event.target.value)
+      newValues.interval_start = newIntervalSizeValue
```
hurrendor, line 209: On line 205 it looks like interval_size is being set, but on line 209 it's being set again conditionally to the same value? Should it be conditionally set?

src/components/pages/collectRecordFormPages/BenthicPitForm/BenthicPitTransectInputs.jsx
```diff
@@ -215,10 +223,21 @@ const BenthicPitTransectInputs = ({
...
+      ...formik.values,
+      is_interval_size_as_start: checked,
     }
+
+    if (checked && currentIntervalSizeValue) {
+      newValues.interval_start = currentIntervalSizeValue
+    }
+
+    formik.setValues(newValues)
+    resetNonObservationFieldValidations({
```
hurrendor, line 238: Is this trying to skip validation or just reset?

hurrendor (changes requested):
Conditional variable setting?

src/components/pages/collectRecordFormPages/BenthicPitForm/BenthicPitTransectInputs.jsx
hurrendor: Ah whoops I misread that. Thanks for clarifying!

PR #1292, opened 2025-07-14
fix: [M1574] failing tests

tienhoah:
## Summary by CodeRabbit

* **New Features**
  * Added test identifiers to key navigation links in the header and to the protocol tooltip in record forms, improving testability.

* **Tests**
  * Updated tests to use new test identifiers for selecting navigation links and protocol tooltips, ensuring more reliable and maintainable test cases.

### DIFF
--- a/src/components/RecordFormTitle/RecordFormTitle.jsx
+++ b/src/components/RecordFormTitle/RecordFormTitle.jsx
@@ -53,7 +53,7 @@ const RecordFormTitle = ({
 }) => {
   const { t } = useTranslation()
   const transectType = getProtocolTransectType(protocol)
-  const protocolTitle = t(`protocol_titles.${protocol}`) ?? ''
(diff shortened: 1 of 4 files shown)


src/components/RecordFormTitle/RecordFormTitle.jsx
```diff
@@ -53,7 +53,7 @@ const RecordFormTitle = ({
 }) => {
   const { t } = useTranslation()
   const transectType = getProtocolTransectType(protocol)
-  const protocolTitle = t(`protocol_titles.${protocol}`) ?? ''
+  const protocolTitle = t(`protocol_titles.${protocol}`, '')
```
hurrendor, line 56: Is the empty string necessary?

hurrendor (changes requested):
I'm uncertain about the empty string in the translate function. Has this been tested and still populates without error?

PR #1294, opened 2025-07-17
M1555 Re-add link to attribute proposal

hurrendor:
## Summary by CodeRabbit

* **Enhancements**
  * Improved the "Genus not found" message to include a styled contact link for easier user support.
  * Updated autocomplete dropdown to display the "no results" message only when relevant, reducing empty or unnecessary content.


src/locales/en/translation.json
```diff
@@ -171,7 +171,7 @@
   "taxonomies": {
     "add_fish_species": "Add new fish species",
     "genus": "Genus",
-    "genus_not_found": "Genus not found in MERMAID",
+    "genus_not_found": "<p data-testid='noResults'>Genus not found in MERMAID </p><a>$t(contact_us)</a>",
     "parent": "Parent",
```
tienhoah, line 175: should <a> stay inside <p> ? Either way is fine, if you have intention to make a 2 separate elements.

tienhoah (approved):

src/locales/en/translation.json
hurrendor: This one was intended to have two separate elements - that way the link shows on another line vs being inline.

PR #1295, opened 2025-07-18
M1555 Fix unit tests

hurrendor:
## Summary by CodeRabbit

* **Tests**
  * Improved test coverage for the InputAutocomplete component by distinguishing behavior when the noResultsText prop is present or absent.
  * Added and refined tests to ensure correct display of the "no results" view based on prop usage.


hurrendor:
[Drop down list is not visible when proposing new fish species and BA](https://trello.com/c/W6rnUDzh/1555-drop-down-list-is-not-visible-when-proposing-new-fish-species-and-ba)

PR #1543, opened 2026-03-31
1895 cdk nag

ms280690:
https://trello.com/c/tENy7ffx/1895-add-cdk-nag-security-scanning-to-webapp-iac

## Summary by CodeRabbit

* **New Features**
  * PRs run a new CI workflow: cached deps, lint, tests, infra checks, single compliance report on PRs, and a CDK diff step that can fail on unsuppressed errors.

* **Documentation**
  * Added guidance for infra check findings and how to add/manage suppressions locally; documented synth/report workflow.

* **Tests**
  * Storybook-focused test config and script added; test runner defaults adjusted.

* **Chores**
  * CDK tooling updated, CDK Nag added, and an additional infra issue ID acknowledged.

* **Style**
  * Lint config updated to ignore a new unresolved-import pattern.


saanobhaai (changes requested):
I just can't judge the validity of AI comments without replies/resolutions -- most look not valid but a few seem potentially serious and I can't tell without doing a lot of investigation. Can you please handle these in some fashion?

iac/cdk_diff.sh
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env bash
```
michaelconnor00, line 1: This script should look for an existing comment, and overwrite it with the new diff, instead of creating a comment for each execution. It is confusing and bloat having so many comments.

michaelconnor00 (changes requested):
One comment about the cdk diff script. This change should propagate to the other repos.

iac/cdk_diff.sh
ms280690: For API --> https://github.com/data-mermaid/mermaid-api/pull/673
For Dashboard --> https://github.com/data-mermaid/mermaid-dash-v2/pull/228

michaelconnor00 (approved):

.github/workflows/pr.yml
```diff
@@ -0,0 +1,172 @@
...
+          set -o pipefail
+          npx cdk synth --quiet 2>&1 | tee /tmp/cdk-nag.log
+          if grep -q '^\[Error' /tmp/cdk-nag.log; then
+            echo "has_errors=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "has_errors=false" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Post CDK Nag results to PR
+        if: always() && steps.cdk-nag.outcome == 'success'
```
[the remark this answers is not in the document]
michaelconnor00: @ms280690 have you verified that the workflow runs as expected when cdk-nag finds and issue?

.github/workflows/pr.yml
```diff
@@ -0,0 +1,172 @@
...
+          CDK_DEFAULT_ACCOUNT: '000000000000'
+          CDK_DEFAULT_REGION: us-east-1
+        run: |
+          set -o pipefail
+          npx cdk synth --quiet 2>&1 | tee /tmp/cdk-nag.log
+          if grep -q '^\[Error' /tmp/cdk-nag.log; then
+            echo "has_errors=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "has_errors=false" >> "$GITHUB_OUTPUT"
+          fi
```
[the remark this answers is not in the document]
michaelconnor00: @ms280690  have you verified that the workflow runs as expected when cdk-nag finds and issue?

saanobhaai (approved):

saanobhaai (changes requested):
Bunch of changes to storybook and other code unrelated to cdk-nag

iac/nag-suppressions.ts
```diff
@@ -0,0 +1,97 @@
...
+      reason: `${TODO}: Evaluate adding AWS WAF integration.`,
+    },
+    {
+      id: 'AwsSolutions-CFR3',
+      reason: `${TODO}: Enable CloudFront access logging.`,
+    },
+    {
+      id: 'AwsSolutions-CFR7',
+      reason: `${TODO}: Migrate from Origin Access Identity (OAI) to Origin Access Control (OAC).`,
+    },
```
JBurkinshaw: The static-site construct already uses OAC at `iac/static-site.ts:101` via `cloudfront_origins.S3BucketOrigin.withOriginAccessControl(siteBucket)`. CFR7's check is "the distribution uses OAC for the S3 origin" - if it's truly firing, something about how the S3 origin is wired is making cdk-nag think OAC isn't in place; if it's not firing, this suppression is dead code with a misleading reason. Could you confirm by removing this entry and running `cdk synth` to see whether the rule actually triggers, and if it does, dig into why before adding a TODO that suggests work that's already done?

iac/nag-suppressions.ts
```diff
@@ -0,0 +1,97 @@
...
+
+const ACCEPTED = 'ACCEPTED'
+const TODO = 'TODO'
+
+export function applyNagSuppressions(stack: cdk.Stack): void {
+  NagSuppressions.addStackSuppressions(stack, [
+    // --- S3 ---
+    {
+      id: 'AwsSolutions-S1',
+      reason: `${TODO}: Enable server access logging on the site bucket.`,
```
JBurkinshaw, line 19: I might be missing the point, but are all of these valid TODOs?
If they are, do we have corresponding tickets in trello?

.github/workflows/pr.yml
```diff
@@ -0,0 +1,216 @@
...
+        id: yarn-cache
+        with:
+          path: node_modules
+          key: ${{ runner.os }}-yarn-${{ hashFiles('yarn.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-yarn-
+
+      - if: ${{ steps.yarn-cache.outputs.cache-hit != 'true' }}
+        name: Install dependencies
+        run: yarn install
```
JBurkinshaw: The `deps-cache` job runs `yarn install` (mutable, will rewrite the lockfile if it drifts). Every downstream job uses `--immutable`. So if a PR forgets to commit lockfile changes, the cache job silently fixes it and the rest of CI passes against a different lockfile than what merges.

Is this difference intentional? Could you make `deps-cache` use `--immutable` too?

.github/workflows/pr.yml
```diff
@@ -0,0 +1,216 @@
...
+  id-token: write
+  issues: write
+  pull-requests: write
+
+concurrency:
+  group: pr-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+jobs:
+  deps-cache:
```
JBurkinshaw, line 24: The `deps-cache`, `call-lint` and `call-test` blocks are near-identical to deployment.yml. If we add another linter, change the cache key, or change the environment-selection logic, we now have to remember two files. The lint and test parts are already extracted as reusable `_lint.yml` / `_test.yml`; the dep-cache block could be too, or these two top-level workflows could be unified via `on: { push: ..., pull_request: ... }` with conditional steps.

Can you either extract a `_deps-cache.yml` reusable workflow and have both call it, or consolidate `deployment.yml` and `pr.yml` into a single workflow with event-conditional jobs. The goal would be one place to change cache plumbing.

PR #1554, opened 2026-04-14
Add release script

JBurkinshaw:
Add scripts/release.sh to automate the release process: merges develop
into main, bumps the version via npm version (keeping package.json in
sync with git tags), pushes the tag to trigger production deployment,
back-merges into develop, and creates a GitHub release.

Usage: ./scripts/release.sh <major|minor|patch>

Add a pre-push hook warning when pushing directly to main or develop,
prompting for confirmation. The release script bypasses this via
SKIP_BRANCH_WARNING=1.

---Every PR---

- [ ] Changes have been tested locally
- [ ] Lint has run and any errors have been resolved
- [ ] Prettier has run on any changed files
- [ ] Unit tests have run and passed

---Transitional Changes---

- [ ] touched JS files have been updated with TypeScript
      -- Update where possible and within scope

---Post-Merge---

## Summary by CodeRabbit

* **Chores**
  * Enhanced development workflow with safeguards preventing accidental pushes to protected branches.
  * Added automated release management tooling for streamlined version updates and GitHub release creation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/data-mermaid/mermaid-webapp/pull/1554)


scripts/release.sh
```diff
@@ -0,0 +1,77 @@
...
+NEW_VERSION="v$(node -p "require('./package.json').version")"
+echo "New version: $NEW_VERSION"
+
+# Push main + tag in one go
+# --follow-tags: pushes annotated tags reachable from pushed commits (not all local tags)
+# Tag push triggers production deploy via .github/workflows/deployment.yml
+echo "Pushing main and tag..."
+git push origin main --follow-tags
+
+# Back-merge into develop via PR (branch protection prevents direct push)
```
saanobhaai: The manual path doesn't require this; I'm not getting why it's necessary at all.

scripts/release.sh
```diff
@@ -0,0 +1,77 @@
...
+echo "Pulling latest develop and main..."
+git checkout develop && git pull
+git checkout main && git pull
+
+# Merge develop into main
+# --no-ff: forces a merge commit so releases are visible in history and individually revertable
+echo "Merging develop into main..."
+git merge develop --no-ff -m "Merge develop into main for release"
+
+# Bump version — updates package.json, creates commit, creates annotated tag
```
saanobhaai, line 39: Slightly nervous about tagging logic... Are we sure the tag will always be what we want, with the `v` etc.?

saanobhaai (changes requested):
Looks generally good, but strikes me as not actually reducing human steps much, and introduces another dependency, script, and complexity. Main advantage it seems to me is bumping the version in package.json at the same time. Let's discuss.

scripts/release.sh
JBurkinshaw: Good catch. I just updated it to capture the output of `npm version` directly, which already includes the `v` prefix. `npm version <major|minor|patch>` reads the current version from `package.json`, increments the relevant segment (e.g., `2.31.0` → `2.32.0` for minor), writes it back, and creates a tagged commit (all in one step). That means the tag should always be consistent with what's in `package.json`

scripts/release.sh
JBurkinshaw: Originally the script created a throwaway branch + PR for this step because I didn't have admin push rights to `develop` (branch protection required changes via PR). Since then I've been granted admin rights, so I've updated the script to push directly to match the manual workflow.

scripts/release.sh
```diff
@@ -0,0 +1,63 @@
...
+# Back-merge version bump into develop
+echo "Back-merging main into develop..."
+git checkout develop
+git merge main
+SKIP_BRANCH_WARNING=1 git push origin develop
+
+# Create GitHub release with auto-generated changelog
+# --target main: ensures release creation works even if the tag hasn't propagated yet
+echo "Creating GitHub release..."
+gh release create "$NEW_VERSION" --generate-notes --target main
```
JBurkinshaw, line 60: `--target main` should work around a race condition where GitHub hasn't registered the pushed tag yet. I hit this when I tried the release manually, The `gh release create` command failed saying the tag didn't exist remotely despite just being pushed.

PR #1576, opened 2026-05-23
feat: [M1956] Handle proposed attributes

gwan93:
Ref: https://trello.com/c/AvUboctz/1956-attribute-visibility-filter-hold-referenced-attributes

This PR handles the visibility of proposed attributes for benthic and fish species. A proposed attribute is removed from a user's indexedDB only if it is not being used by any of the user's collect records. If the user has a collect record that uses a proposed attribute, then it will not be removed until all of the user's collect records using that proposed attribute have been deleted or submitted. 

Updates to behavior:
- when api data is pulled, a new check has been added to stop the removal of an attribute if that attribute is being used in one or more of the user's collect records
- when a collect record is deleted or submitted, a new check has been added to see if any other collect records use the deleted/submitted record's attribute. If no other collect records use it, then that attribute is checked against the api to see if it has a proposed status. If so, then it is removed from the user's indexedDB

---Every PR---

---Transitional Changes---

---Post-Merge---

## Release Notes

* **New Features**
  * Added attribute usage tracking to monitor which benthic and fish species attributes are referenced in your data.
  * Automatic cleanup of inaccessible attributes when submitting or deleting sample units.

* **Bug Fixes**
  * Improved handling of proposed attributes to prevent unnecessary storage of unused references.

* **Localization**
  * Added error message for unavailable attributes.

[Review Change Stack](https://app.coderabbit.ai/change-stack/data-mermaid/mermaid-webapp/pull/1576?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


src/components/pages/collectRecordFormPages/CollectRecordFormPage/CollectRecordFormPage.jsx
```diff
@@ -438,6 +444,79 @@ const CollectRecordFormPage = ({
...
+              }
+            }
+          })
+        }),
+      )
+    } catch {
+      toast.error(...getToastArguments(t('sample_units.errors.attribute_unavailable')))
+    }
+  }
+
```
tienhoah: **Issue:** `removeInaccessibleAttributes` bypasses the DatabaseSwitchboard abstraction
`CollectRecordFormPage.jsx` currently accesses `databaseSwitchboardInstance._dexiePerUserDataInstance` directly (lines 449, 497) and makes raw axios API calls. Per our architecture rules, component code should never touch IDB or the API directly — all of that goes through the switchboard.

**To do:**
Move `removeInaccessibleAttributes` out of `CollectRecordFormPage.jsx` and into `BenthicAttributesMixin.js` as a switchboard method. The switchboard already has everything it needs:
- `this._dexiePerUserDataInstance` — for IDB reads/deletes
- `this._getAccessToken` — for auth (same pattern used in SitesMixin, GfcrMixin, SubmittedRecordsMixin)
- `this._apiBaseUrl` — for the API URL

In the component, replace the inline function + its 5 imports (axios, getAuthorizationHeaders, useAuthentication, dexieCurrentUserInstance, getAttributesInUse) and the getAccessToken hook call with a single call:
`databaseSwitchboardInstance.removeInaccessibleAttributes(collectRecordBeingEdited.data).catch(() => {})`

Also fix the axios call inside the moved method — it's currently:
axios.get(url, authHeaders)  // ❌ auth headers never sent
Should be:
axios.get(url, { headers: authHeaders })  // ✅
(This is also a bug on its own — without this fix, the API call silently sends no auth and will get a 401.)

src/components/pages/collectRecordFormPages/CollectRecordFormPage/CollectRecordFormPage.jsx
```diff
@@ -490,6 +570,7 @@ const CollectRecordFormPage = ({
         setIsDeletingRecord(false)
         toast.success(...getToastArguments(t('sample_units.success.record_deleted')))
         navigate(`${ensureTrailingSlash(currentProjectPath)}collecting/`)
+        removeInaccessibleAttributes(collectRecordBeingEdited.data).catch(() => {})
```
tienhoah: **Issue:** The .catch(() => {}) on the call site swallows all errors including the toast inside `removeInaccessibleAttributes` — so the error message in the i18n file will never be shown. Either remove the inner `toast.error` (since errors are silently swallowed anyway), or restructure so the toast fires before navigation. Also add a null guard on `collectRecordBeingEdited?.data` before passing it in.

**Potential fix:**
After refactoring `removeInaccessibleAttributes` to the mixin, the simplest fix is to flip the order — run `removeInaccessibleAttributes` first, show the toast if it fails, then navigate regardless:

```
databaseSwitchboardInstance.removeInaccessibleAttributes(collectRecordBeingEdited.data)
  .catch(() => {
    toast.error(...getToastArguments(t('sample_units.errors.attribute_unavailable')))
  })
  .finally(() => {
    navigate(`${ensureTrailingSlash(currentProjectPath)}collecting/`)
  })
```

This way:
- The toast fires before navigation (while the component is still mounted)
- Navigation always happens whether the cleanup succeeded or failed
- The inner toast.error inside `removeInaccessibleAttributes` can be removed since error handling is now the caller's responsibility

src/App/mermaidData/getAttributesInUse.ts
```diff
@@ -0,0 +1,68 @@
...
+
+interface DexiePerUserDataInstance {
+  collect_records: {
+    toArray: () => Promise<CollectRecord[]>
+  }
+}
+
+// Looks at all fish species and benthic attributes used in collect records,
+// grouped by attribute type with counts per attribute ID.
+const getAttributesInUse = async (
```
tienhoah, line 27: **Question:**
Can invertebrate attributes have a proposed status (status 10)? If so, `getAttributesInUse.ts` needs to handle `obs_belt_inverts` with invertebrate_attribute — it's currently not tracked, so proposed invertebrate attributes would never be retained during sync or cleaned up after submit/delete. If they can't be proposed, add a comment explaining why it's intentionally excluded.

tienhoah (changes requested):
We are getting there!

Another worth considering for this minor follow-up about the `GatedFeature.tsx` and `getCurrentUserOptionalFeature`: 

`getCurrentUserOptionalFeature` return type could be improved to OptionalFeature | undefined instead of Partial<OptionalFeature> — the ?? {} fallback in `GatedFeature.tsx` becomes dead code since the function never returns nullish. Worth a small cleanup PR.

Also, could you rename `index.js` to `index.ts`? We try to move to TypeScript whenever possible, and it’s especially required for new code.

src/locales/id/translation.json
```diff
@@ -875,6 +875,7 @@
       "submitted_moved_to_collecting": "Catatan yang dikirimkan telah dipindahkan ke pengumpulan.",
       "submitted_not_editable": "Catatan yang dikirimkan belum dapat diedit",
       "supporting_data_unavailable": "Data pendukung untuk membuat unit sampel saat ini tidak tersedia",
+      "attribute_unavailable": "Gagal mengambil detail atribut. Silakan coba lagi.",
```
tienhoah: Remove translation for other language, these will be handled on QA side. We only add new language token and English translation

tienhoah (changes requested):

src/App/mermaidData/getAttributesInUse.ts
gwan93: Kim said we don't need to handle the case for macro invertebrates since that work will be implemented after this gets merged. This work is mainly to just removed inaccessible attributes that exist already on users' local indexeddb

tienhoah (approved):

PR #1577, opened 2026-05-26
feat: [M1945] Offline sync for invert_attributes

gwan93:
Ref: https://trello.com/c/ZkHgJuaR/1945-idea-add-offline-sync-for-the-invertattributes-taxonomy-phase-12

This PR adds some fundation work for invert_attributes. All the requirements within the trello card have been met

---Every PR---

---Transitional Changes---

---Post-Merge---

## Summary by CodeRabbit

* **New Features**
  * Added offline storage and synchronization support for invert attributes
  * Implemented methods to retrieve and add invert attributes

* **Tests & Data**
  * Updated test coverage and mock data to support the new invert attributes functionality

[Review Change Stack](https://app.coderabbit.ai/change-stack/data-mermaid/mermaid-webapp/pull/1577?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


src/App/mermaidData/databaseSwitchboard/InvertAttributesMixin.ts
```diff
@@ -0,0 +1,116 @@
...
+  _dexiePerUserDataInstance: DexiePerUserDataInstance
+  _apiSyncInstance: ApiSyncInstance
+  _notAuthenticatedAndReadyError: Error
+  getInvertAttributes(): Promise<InvertAttribute[]>
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type Constructor = new (...args: any[]) => object
+
+const InvertAttributesMixin = (Base: Constructor) =>
```
tienhoah: Since this is the first mixin written in TypeScript while all others are still JS, the local interface definitions (e.g. ApiSyncInstance, DexiePerUserDataInstance) can't be validated against the real implementations — which is how the pushThenPullFishOrInvertAttributes bug slipped through.

I'd recommend converting this back to .js for now to keep it consistent with the other mixins. The value of TypeScript here is limited until the surrounding code is also typed. A dedicated ticket to migrate all mixins together would be the right time to define shared interfaces that the compiler can actually enforce across the whole mixin chain.

Apologize for saying we need move forward to all ts.

src/App/mermaidData/databaseSwitchboard/InvertAttributesMixin.ts
```diff
@@ -0,0 +1,116 @@
...
+        parent: invertAttributeParentId,
+        uiState_pushToApi: true,
+      }
+
+      if (this._isOnlineAuthenticatedAndReady) {
+        const _protectAgainstNetworkStutter =
+          await this._dexiePerUserDataInstance.invert_attributes.put(newInvertAttributeObject)
+
+        return this._apiSyncInstance
+          .pushThenPullFishOrInvertAttributes('invert_attributes')
```
tienhoah: `pushThenPullFishOrInvertAttributes` is called here but this method doesn't exist on SyncApiDataIntoOfflineStorage — any user adding an invert attribute while online will hit a runtime crash.

Does the new protocol work the same way as `Fish or Benthic`? If so, the existing `pushThenPullFishOrBenthicAttributes` is already generic and accepts any data type string, so no new method is needed. Just call that instead:

```
return this._apiSyncInstance
  .pushThenPullFishOrBenthicAttributes('invert_attributes')
  .then((response) => {
    const newInvertAttributeFromApi = response.data.invert_attributes.updates[0]
    return newInvertAttributeFromApi
  })
```

src/App/mermaidData/pullApiData.js
```diff
@@ -68,6 +68,7 @@ export const pullApiData = async ({
     await dexiePerUserDataInstance.transaction(
       'rw',
       dexiePerUserDataInstance.benthic_attributes,
+      dexiePerUserDataInstance.invert_attributes,
```
tienhoah, line 71: `src/App/mermaidData/pullApiData.js` — line **113**

Proposed invert attributes (status: 10) are not filtered out of IndexedDB — unlike `benthic_attributes` and `fish_species`. The test confirms this is the current behaviour (all 3 records including the proposed one are stored).

Can you confirm this is intentional? If invert attributes can also be in a proposed state and should be handled the same way, `invert_attributes` needs to be included in the `isBenthicOrFishSpecies` check. If the difference is a deliberate product decision, a short comment in `pullApiData.js` would help prevent someone from "fixing" it later by accident.

tienhoah (changes requested):
**Other change:**

`src/App/mermaidData/databaseSwitchboard/BenthicAttributesMixin.js` — line **22**

Pre-existing bug spotted while reviewing — the guard clause is missing a return, so the rejection is silently discarded and execution continues:

// missing return
`Promise.reject(new Error('addBenthicAttribute was implemented with missing parameters'))`

// should be
`return Promise.reject(new Error('addBenthicAttribute was implemented with missing parameters'))`

This was my mistake doing this 3 years ago, not blocking this PR but worth fixing while you're in the area.

tienhoah (approved):

PR #1578, opened 2026-05-27
feat: [M1956] Fix regressions: attributes unintentionally removed

gwan93:
Ref: https://trello.com/c/AvUboctz/1956-attribute-visibility-filter-hold-referenced-attributes

This fixes some bugs [introduced previously](https://github.com/data-mermaid/mermaid-webapp/pull/1576) regarding attributes being unintentionally removed from the user's indexeddb when they proposed new attributes.

Bug:
- When a user proposed a new attribute, it would cause a push to the api to create the new attribute. Afterwards a pull would be performed and the attribute would end up in the `updates` array. The previous implementation removed any attributes in the updates array with a status of 10 (proposed)

Changes:
- Removed previous changes to the updates array
- Removed function that would manually remove attributes locally. This is all handled by the api now

---Every PR---

---Transitional Changes---

---Post-Merge---

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed offline sync to retain all benthic attributes on initial load.

* **Refactor**
  * Persist "proposed" records from the API during sync instead of auto-filtering them.
  * Removed automatic cleanup of unavailable attributes, preventing related error toasts and enabling immediate navigation after submit/delete.

[Review Change Stack](https://app.coderabbit.ai/change-stack/data-mermaid/mermaid-webapp/pull/1578?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


tienhoah (changes requested):
@gwan93  Look good, found these changes not really related to this pr, but I think its still worth to tackle.

1) Nit pick: `pushThenPullFishOrBenthicAttributes` is actually fully generic (just passes the string to `pullApiData`), so calling it with `'invert_attributes'` works fine. But the name is **misleading** now that it handles a third type — might be worth renaming to `pushThenPullAttributes` at some point to avoid confusion for future readers. Not a blocker.

---

2) Quick question on `getAttributesInUse.ts` line **57** — was it intentional to include pulled records that have no id?
`(record) => record?.id === undefined || !localRecordIds.has(record.id)`. The way I read the intent: this filter is meant to find pulled records that aren't already in local IndexedDB, so we don't double-count them. An ID-less record can't be deduplicated against local records at all, so including it seems like it would inflate the attribute usage count unexpectedly. I'd expect `&&` here instead of `||`. Am I missing something?

tienhoah (approved):

PR #1582, opened 2026-05-29
feat [M1951] update methods filter with invert

gwan93:
Ref: https://trello.com/c/Py5MZq9o/1951-idea-register-macroinvertebrate-in-the-methods-filter-dropdown-phase-24

Updated the methods filter to now include macro invertebrates. Only users with the optional feature of macroinvertebrate_enabled can see this. The GatedFeature pattern cannot be reused here unfortunately so getting macrovertebrate_enabled has to be obtained using getUserOptionalFeature.

The implementation may look weird: adding macroinvertebrate to the array and then filtering it back out - but this is so that it's easier to remove later on

image

---Every PR---

---Transitional Changes---

---Post-Merge---


src/components/MethodsFilterDropDown/MethodsFilterDropDown.tsx
```diff
@@ -47,20 +56,32 @@ const methodKeys = [
...
+}: MethodsFilterDropDownProps) => {
   const { t } = useTranslation()
-  const methods = methodKeys.map((key) => t(key))
 
-  const handleChange = (event) => {
-    const eventValue = event.target.value
-    handleMethodsColumnFilterChange(eventValue)
+  // Remove the following code when macroinvertebrate goes live
+  const { currentUser } = useCurrentUser()
+  const { enabled: isMacroinvertebrateEnabled } = getCurrentUserOptionalFeature(
```
tienhoah: `const { enabled: isMacroinvertebrateEnabled } = getCurrentUserOptionalFeature(...)`
Every other call to `getCurrentUserOptionalFeature` in the codebase destructures `enabled` with = false as a default (e.g. `Projects.tsx`, `ProjectToolBarSection.tsx`). Without it, `isMacroinvertebrateEnabled` is typed `boolean | undefined` instead of `boolean`. The ternary still works since `undefined` is falsy, but it diverges from the established pattern. Can you add the default?

`const { enabled: isMacroinvertebrateEnabled = false } = getCurrentUserOptionalFeature(...)`

src/components/MethodsFilterDropDown/MethodsFilterDropDown.tsx
```diff
@@ -47,20 +56,32 @@ const methodKeys = [
...
+  // Remove down to here
+
+  const methods = isMacroinvertebrateEnabled
+    ? methodKeys.map((key) => t(key))
+    : // Remove false case when macroinvertebrate goes live
+      methodKeys.filter((key) => key !== 'protocol_titles.macroinvertebrate').map((key) => t(key))
+
+  const handleChange = (event: SelectChangeEvent<string[]>) => {
+    handleMethodsColumnFilterChange(event.target.value as string[])
   }
```
tienhoah, line 86: The "remove when live" comments are helpful, but there are three of them scattered across ~16 lines — the inline one on the ternary branch is easy to miss during cleanup. This is non blocking, just flag it that make the note yourself to clean up later

tienhoah (changes requested):

tienhoah (approved):

PR #1587, opened 2026-06-01
feat [M1954] - Add macroinvert form partially

gwan93:
Ref: https://trello.com/c/087Kh6W4/1954-idea-macroinvertebrate-collect-form-scaffold-routing-transect-inputs-phase-31

Added the collect record form for macro invertebrates.
The urls will use "macroinvertebrate" instead of "beltinvert" as the ticket originally suggests.
A lot of logic is recycled from the fish belt form

---Every PR---

---Transitional Changes---

---Post-Merge---

## New Features

* Added macroinvertebrate belt invert collection capability to the app
* Users can create and edit macroinvertebrate collection records with comprehensive transect inputs (depth, width, slope, tide, visibility, and others)
* Includes observation table for recording invertebrate observations
* New macroinvertebrate option available in the collecting dropdown menu
* Feature is controlled by a macroinvertebrate flag

### DIFF
--- a/src/components/pages/collectRecordFormPages/BeltInvertForm/BeltInvertForm.tsx
+++ b/src/components/pages/collectRecordFormPages/BeltInvertForm/BeltInvertForm.tsx
@@ -1,0 +1,189 @@
+import { toast } from 'react-toastify'
+import { useParams } from 'react-router'
+import React, { useState, useEffect, useCallback, useReducer, useMemo } from 'react'
+import { useTranslation } from 'react-i18next'
+
+import {
+  getBeltInvertAdditionalValues,
+  getCollectRecordDataInitialValues,
+  getSampleInfoInitialValues,
+  getTransectInitialValues,
+} from '../collectRecordFormInitialValues'
+import { getDataForSubNavNode } from '../../../../library/getDataForSubNavNode'
+import { getToastArguments } from '../../../../library/getToastArguments'
+import { reformatFormValuesIntoBeltInvertRecord } from '../reformatFormValuesIntoRecord'
+import { sortArrayByObjectKey } from '../../../../library/arrays/sortArrayByObjectKey'
+import { useCurrentUser } from '../../../../App/CurrentUserContext'
+import { useDatabaseSwitchboardInstance } from '../../../../App/mermaidData/databaseSwitchboard/DatabaseSwitchboardContext'
+import { useHttpResponseErrorHandler } from '../../../../App/HttpResponseErrorHandlerContext'
+import { useSyncStatus } from '../../../../App/mermaidData/syncApiDataIntoOfflineStorage/SyncStatusContext'
+import { useUnsavedDirtyFormDataUtilities } from '../../../../library/useUnsavedDirtyFormDataUtilities'
+import CollectRecordFormPage from '../CollectRecordFormPage'
+import ErrorBoundary from '../../../ErrorBoundary'
+import beltInvertObservationReducer from './beltInvertObservationReducer'
+import BeltInvertObservationTable from './BeltInvertObservationTable'
+import BeltInvertTransectInputs from './BeltInvertTransectInputs'
+import useIsMounted from '../../../../library/useIsMounted'
+
+interface BeltInvertFormProps {
+  isNewRecord?: boolean
+}
+
+const BeltInvertForm = ({ isNewRecord = true }: BeltInvertFormProps) => {
+  const { t } = useTranslation()
+  const { recordId, projectId } = useParams()
+  const { isSyncInProgress } = useSyncStatus()
+  const isMounted = useIsMounted()
+
+  const { databaseSwitchboardInstance } = useDatabaseSwitchboardInstance()
+  const handleHttpResponseError = useHttpResponseErrorHandler()
+  const observationsReducer = useReducer(beltInvertObservationReducer, [])
+  const [areObservationsInputsDirty, setAreObservationsInputsDirty] = useState(false)
+  const [collectRecordBeingEdited, setCollectRecordBeingEdited] = useState()
+  const [idsNotAssociatedWithData, setIdsNotAssociatedWithData] = useState([])
+  const [isLoading, setIsLoading] = useState(true)
+  const [observationsState, observationsDispatch] = observationsReducer
+  const [sites, setSites] = useState([])
+  const [subNavNode, setSubNavNode] = useState(null)
+  const [invertAttributes, setInvertAttributes] = useState([]) // to be used in phase 3.2
+  const { currentUser } = useCurrentUser()
+
+  const errorMessage = isNewRecord
+    ? t('sample_units.errors.supporting_data_unavailable')
+    : t('sample_units.errors.data_unavailable')
+
+  const _getSupportingData = useEffect(() => {
+    if (databaseSwitchboardInstance && projectId && !isSyncInProgress) {
+      const promises = [
+        databaseSwitchboardInstance.getSitesWithoutOfflineDeleted(projectId),
+        databaseSwitchboardInstance.getInvertAttributes(),
+        databaseSwitchboardInstance.getProject(projectId),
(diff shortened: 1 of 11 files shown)


src/components/pages/collectRecordFormPages/BeltInvertForm/beltInvertObservationReducer.js
```diff
@@ -0,0 +1,88 @@
+import { createUuid } from '../../../../library/createUuid'
+
+const beltInvertObservationReducer = (state, action) => {
+  switch (action.type) {
+    case 'loadObservationsFromApi': {
+      const updateObservationsWithIds = [...action.payload].map((record) => ({
+        id: record.id || createUuid(),
+        ...record,
+      }))
```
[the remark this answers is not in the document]
gwan93: Will not implement. Other forms follow the same pattern

src/components/pages/collectRecordFormPages/BeltInvertForm/BeltInvertForm.tsx
```diff
@@ -0,0 +1,189 @@
...
+          },
+        )
+        .catch((error) => {
+          handleHttpResponseError({
+            error,
+            callback: () => {
+              toast.error(...getToastArguments(errorMessage))
+            },
+          })
+        })
```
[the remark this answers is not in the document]
gwan93: Will not implement. Other forms follow the same pattern

src/components/pages/collectRecordFormPages/BeltInvertForm/BeltInvertTransectInputs.tsx
```diff
@@ -0,0 +1,447 @@
...
+          sizeBinValidationProperties,
+          'size_bin',
+        )}
+        onBlur={formik.handleBlur}
+        value={formik.values.size_bin as string}
+        onChange={(event: React.ChangeEvent<HTMLElement>) =>
+          handleInputChange(event, 'size_bin', SIZE_BIN_VALIDATION_PATH)
+        }
+        helperText={t('fish_size_bin_info')}
+      />
```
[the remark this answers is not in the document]
gwan93: Does not need to be changed. Original ticket says "size_bin reads its options from the API's invertsizebins choices endpoint; width reads from invertbelttransectwidths (both confirmed merged on mermaid-api/dev in api/resources/choices.py:50-83). UX matches FishBelt's fishsizebins / belttransectwidths controls; only the choices source differs. "

src/components/pages/collectRecordFormPages/BeltInvertForm/beltInvertObservationReducer.ts
```diff
@@ -0,0 +1,88 @@
...
+        const newCountToUse = Number.isNaN(parsedCount) || newCount === '' ? null : parsedCount
+
+        return isObservationToUpdate ? { ...observation, count: newCountToUse } : observation
+      })
+    case 'updateSize':
+      return state.map((observation) => {
+        const isObservationToUpdate = observation.id === action.payload.observationId
+        const { newSize } = action.payload
+
+        const newSizeToUse = Number.isNaN(newSize) || newSize === '' ? null : parseFloat(newSize)
```
tienhoah: `src/components/pages/collectRecordFormPages/BeltInvertForm/beltInvertObservationReducer.ts` — `updateSize`

`Number.isNaN(newSize)` will never be `true` when `newSize` is a string — it only matches the actual `NaN` value, not un-parseable input like `"abc"`. That means a user typing invalid text stores `NaN` in state instead of `null`. The `updateCount` case was fixed in the "Improve count logic" commit to parse first; `updateSize` needs the same treatment:

```ts
const parsedSize = parseFloat(newSize)
const newSizeToUse = Number.isNaN(parsedSize) || newSize === '' ? null : parsedSize

File: src/components/pages/collectRecordFormPages/BeltInvertForm/beltInvertObservationReducer.js [hunk 5]
```diff
@@ -0,0 +1,88 @@
+import { createUuid } from '../../../../library/createUuid'
+
+const beltInvertObservationReducer = (state, action) => {
```
tienhoah, line 3: **[Question]** `src/components/pages/collectRecordFormPages/BeltInvertForm/beltInvertObservationReducer.ts`

Just double check about this reducer, which is following FishBelt line-for-line as the ticket instructs — does the macroinvertebrate API also use `fish_attribute` as the observation attribute field name, or is it something different (e.g. `invert_attribute`)? Worth confirming before Phase 3.2 wires up the real observations table.

src/components/pages/collectRecordFormPages/BeltInvertForm/BeltInvertForm.tsx
```diff
@@ -0,0 +1,189 @@
...
+  ])
+
+  const { getPersistedUnsavedFormData: getPersistedUnsavedFormikData } =
+    useUnsavedDirtyFormDataUtilities(`${currentUser.id}-unsavedSampleInfoInputs`)
+
+  const initialFormikFormValues = useMemo(() => {
+    const collectRecordInitialValues = getPersistedUnsavedFormikData() ?? {
+      ...getCollectRecordDataInitialValues(collectRecordBeingEdited),
+      ...getSampleInfoInitialValues(collectRecordBeingEdited),
+      ...getBeltInvertAdditionalValues(collectRecordBeingEdited),
```
tienhoah: `src/components/pages/collectRecordFormPages/BeltInvertForm/BeltInvertForm.tsx` — `getBeltInvertAdditionalValues`

The ticket says to add `getBeltInvertAdditionalValues` mirroring `getFishBeltAdditionalValues`, but `getFishBeltAdditionalValues` doesn't exist in the codebase — and `FishBeltForm` itself doesn't use one. It calls `getTransectInitialValues('fishbelt_transect')` directly, which already covers all the transect fields. `getBeltInvertAdditionalValues` returns the exact same fields as `getTransectInitialValues('beltinvert_transect')`, so the spread in `collectRecordInitialValues` is silently overwritten. Can we drop `getBeltInvertAdditionalValues` and follow the FishBelt pattern exactly?

src/components/pages/collectRecordFormPages/BeltInvertForm/beltInvertObservationReducer.js
tienhoah, line 3: `src/components/pages/collectRecordFormPages/BeltInvertForm/beltInvertObservationReducer.ts`

`state` and `action` are both implicitly `any`. If keeping this as `.ts`, could you add the types in `src/types/collect.ts`? That said, given the other reducers in the codebase are `.js`, it's also fine to just rename this to `.js` to stay consistent.

src/components/pages/collectRecordFormPages/BeltInvertForm/BeltInvertForm.tsx
```diff
@@ -0,0 +1,189 @@
...
+    [invertAttributes],
+  )
+
+  const PartiallyAppliedBeltInvertTransectInputs = useCallback(
+    (props) => {
+      return (
+        <BeltInvertTransectInputs
+          observationsState={observationsState}
+          observationsDispatch={observationsDispatch}
+          {...props}
```
tienhoah, line 157: `src/components/pages/collectRecordFormPages/BeltInvertForm/BeltInvertTransectInputs.tsx`

`observationsState` and `observationsDispatch` are passed in via `PartiallyAppliedBeltInvertTransectInputs` but aren't declared in `BeltInvertTransectInputsProps` and aren't used anywhere in the component. Looking at `FishBeltTransectInputs`, these are used to dispatch `resetFishSizes` when the size bin changes and to check whether observations exist before doing so. That logic is missing here — changing the size bin won't reset observation sizes.

tienhoah (changes requested):

tienhoah (approved):

PR #1597, opened 2026-06-09
fix [M1511] Align filter helper text tooltip with icon and handle off-screen

tienhoah:
## What changed

`FilterSearchToolbar` previously positioned the tooltip using hardcoded pixel offsets (`pixelAdjustTop = 302`, `pixelAdjustLeft = 487`) subtracted from viewport coordinates. These constants were tuned for one specific page layout and produced misaligned tooltips on every other page (Collecting, Submitted, Sites, Management Regimes, etc.).

The fix replaces the magic constants with ref-based measurement: the icon button and label container are measured after the tooltip mounts, the tooltip is centered on the icon relative to its positioned ancestor, and the arrow position tracks the icon even when the box is clamped to the viewport edge. The `type` prop and per-modal overrides (`copy-site-modal`, `copy-mr-modal`) are removed as they're no longer needed.

**Files changed:**
- `FilterSearchToolbar.jsx` — ref-based positioning, viewport clamping, `type` prop removed; now uses `useTooltipPosition` hook
- `ColumnHeaderToolTip.jsx` — `arrowOffset` prop added to `clip-path`, `useLayoutEffect` for flicker-free first render
- `CopySitesModal.jsx` / `CopyManagementRegimesModal.jsx` — removed now-unnecessary `type` prop, reduced modal height so tooltip doesn't get clipped at the top
- `LabelWithTooltip.jsx` (new) — reusable column header component with built-in info tooltip; replaces the old manual `isHelperTextShowing` + `handleInfoIconClick` pattern used in Users page and all 6 observation tables
- `useTooltipPosition.js` (new, in `src/library/`) — shared hook extracted from the duplicated positioning logic in `FilterSearchToolbar` and `LabelWithTooltip`
- Users page and all 6 observation tables (`FishBelt`, `BenthicLit`, `BenthicPit`, `BenthicPhotoQuadrat`, `HabitatComplexity`, `PercentCover`) — migrated to `LabelWithTooltip`; one-at-a-time tooltip behaviour via shared `groupRef`

### Pages to check (click ⓘ next to the filter input):

- Users — Filter by name or email
- Collecting — Filter by site, management, or observer
- Submitted — Filter by site, management, or observer
- Sites — Filter by name, reef type, reef zone, or exposure
- Management Regimes — Filter by name or year
- SU / Observers (Sample Units) — Filter input
- SU / MRs (Sample Units) — Filter input
- GFCR — Filter by title or date

Modals to check (open via Copy button):

- Copy Sites modal — Filter by name, project, or country
- Copy Management Regimes modal — Filter by name, project, or year

Column header tooltips to check (click ⓘ in table headers):

- Users page — Admin / Collector / Read-only columns (only one tooltip open at a time)
- Observation tables — Benthic Attribute, Growth Form, Fish Name, Fish Size, Fish Count, Quadrat Number, Number of Points columns

For each, verify:
- Tooltip arrow aligns with the ⓘ icon
- Tooltip appears above the input/header
- Clicking outside dismisses it
- On a narrow window, tooltip shifts horizontally but arrow still points at the icon
- Only one tooltip is open at a time within the same table

---Every PR---

---Transitional Changes---

---Post-Merge---

## Summary by CodeRabbit

* **New Features**
  * New unified header tooltip and label-with-tooltip components for consistent info display.
* **Improvements**
  * Improved tooltip positioning (arrow alignment and viewport clamping) so tooltips stay on-screen.
  * Only one tooltip in a group is shown at a time for cleaner UI.
  * Modals can now set a configurable maximum height.
* **Refactor**
  * Replaced legacy helper-text/info-icon behavior across many tables/forms with the shared tooltip system.

### DIFF
--- a/src/components/ColumnHeaderToolTip/ColumnHeaderToolTip.tsx
+++ b/src/components/ColumnHeaderToolTip/ColumnHeaderToolTip.tsx
@@ -70,7 +70,6 @@ const ColumnHeaderToolTip = forwardRef<HTMLSpanElement, ColumnHeaderToolTipProps
       <TooltipPopup
         ref={ref}
         role="tooltip"
-        aria-labelledby="tooltip"
         $left={left}
         $maxWidth={maxWidth}
         $top={top}
--- a/src/components/ColumnHeaderToolTip/ColumnHeaderToolTip.tsx
+++ b/src/components/ColumnHeaderToolTip/ColumnHeaderToolTip.tsx
@@ -1,12 +1,18 @@
 import React, { forwardRef, useLayoutEffect } from 'react'
-import PropTypes from 'prop-types'
 import { styled } from 'styled-components'
 import domPurify from 'dompurify'
 import theme from '../../theme'
 
-export const TooltipPopup = styled('span')`
+interface TooltipPopupProps {
+  $maxWidth?: string
+  $left?: string
+  $top?: string
+  $arrowOffset?: string
+}
+
+export const TooltipPopup = styled('span')<TooltipPopupProps>`
   display: block;
-  max-width: ${(props) => props.maxWidth || '25rem'};
+  max-width: min(${(props) => props.$maxWidth || '25rem'}, calc(100vw - 2rem));
   width: max-content;
   background: ${theme.color.primaryColor};
   color: ${theme.color.white};
@@ -23,15 +29,24 @@ export const TooltipPopup = styled('span')`
   );
   padding: 1em;
   padding-bottom: calc(1rem + 15px);
-  left: ${(props) => props.left || '0em'};
-  top: ${(props) => props.top || '0em'};
+  left: ${(props) => props.$left || '0em'};
+  top: ${(props) => props.$top || '0em'};
   white-space: normal;
   z-index: 100;
   text-align: left;
 `
 
-// eslint-disable-next-line react/display-name
-const ColumnHeaderToolTip = forwardRef(
+interface ColumnHeaderToolTipProps {
+  helperText?: React.ReactNode
+  left?: string
+  top?: string
+  maxWidth?: string
+  html?: string
+  arrowOffset?: string
+  onMount?: () => void
+}
+
+const ColumnHeaderToolTip = forwardRef<HTMLSpanElement, ColumnHeaderToolTipProps>(
   (
     {
       helperText = '',
@@ -56,9 +71,9 @@ const ColumnHeaderToolTip = forwardRef(
         ref={ref}
         role="tooltip"
         aria-labelledby="tooltip"
-        left={left}
-        maxWidth={maxWidth}
-        top={top}
+        $left={left}
+        $maxWidth={maxWidth}
+        $top={top}
         $arrowOffset={arrowOffset}
       >
         {/* eslint-disable-next-line react/no-danger */}
@@ -68,14 +83,6 @@ const ColumnHeaderToolTip = forwardRef(
   },
 )
 
-export default ColumnHeaderToolTip
+ColumnHeaderToolTip.displayName = 'ColumnHeaderToolTip'
 
-ColumnHeaderToolTip.propTypes = {
-  helperText: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
-  left: PropTypes.string,
-  top: PropTypes.string,
-  maxWidth: PropTypes.string,
-  html: PropTypes.string,
-  arrowOffset: PropTypes.string,
-  onMount: PropTypes.func,
-}
+export default ColumnHeaderToolTip
--- a/src/components/ColumnHeaderToolTip/LabelWithTooltip.jsx
+++ b/src/components/ColumnHeaderToolTip/LabelWithTooltip.jsx
@@ -1,0 +1,81 @@
+import React, { useState, useEffect, useRef } from 'react'
+import PropTypes from 'prop-types'
+import { LabelContainer } from '../generic/form'
+import { IconInfo } from '../icons'
+import { IconButton } from '../generic/buttons'
+import ColumnHeaderToolTip from './ColumnHeaderToolTip'
+import useTooltipPosition from '../../library/useTooltipPosition'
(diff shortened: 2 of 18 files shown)


alanjleonard (changes requested):
Collector on the users page is too high. 
The rest look good.
Image

src/components/ColumnHeaderToolTip/LabelWithTooltip.jsx
```diff
@@ -0,0 +1,81 @@
...
+import { LabelContainer } from '../generic/form'
+import { IconInfo } from '../icons'
+import { IconButton } from '../generic/buttons'
+import ColumnHeaderToolTip from './ColumnHeaderToolTip'
+import useTooltipPosition from '../../library/useTooltipPosition'
+
+// groupRef is an optional shared ref({ current: closeFn | null }) passed to a set of sibling
+// instances. Opening one tooltip calls groupRef.current() to close whichever is currently open,
+// ensuring only one is visible at a time without lifting state to the parent.
+const LabelWithTooltip = ({ label, tooltipText, groupRef = undefined }) => {
```
gwan93: Can this be converted to typescript?

src/library/useTooltipPosition.js
```diff
@@ -0,0 +1,39 @@
+import { useState, useCallback } from 'react'
+
+const VIEWPORT_MARGIN = 8
+
+const useTooltipPosition = (iconRef, containerRef, tooltipRef) => {
```
gwan93: This can also be converted to typescript

src/components/ColumnHeaderToolTip/LabelWithTooltip.tsx
```diff
@@ -0,0 +1,81 @@
...
+    const handleClickOutside = (event) => {
+      if (containerRef.current && !containerRef.current.contains(event.target)) {
+        setIsShowing(false)
+        if (groupRef) {
+          groupRef.current = null
+        }
+      }
+    }
+
+    document.addEventListener('click', handleClickOutside)
```
gwan93, line 43: This adds the document event listener even when the LabelWithTooltip is closed. It would be better if the listener was added only when then tooltip is opened so we don't have extra document event listeners

src/components/ColumnHeaderToolTip/LabelWithTooltip.jsx
```diff
@@ -0,0 +1,81 @@
...
+        />
+      ) : null}
+    </LabelContainer>
+  )
+}
+
+LabelWithTooltip.propTypes = {
+  label: PropTypes.node.isRequired,
+  tooltipText: PropTypes.node.isRequired,
+  groupRef: PropTypes.shape({ current: PropTypes.func }),
```
gwan93: Sometimes groupRef.current can be null too

gwan93 (changes requested):
Would it be possible to `align-items: flex-start` on the parent div so that there is less gap between the tooltip and the icon?
Image

If the user has a really narrow screen the tooltip seems to get cut off. @alanjleonard do you have any thoughts on this? If none of our users use a narrow screen then this isn't an issue
Image

tienhoah:
> Would it be possible to `align-items: flex-start` on the parent div so that there is less gap between the tooltip and the icon? Image
> 
> If the user has a really narrow screen the tooltip seems to get cut off. @alanjleonard do you have any thoughts on this? If none of our users use a narrow screen then this isn't an issue Image

1) use align-items: flex-start now
2) I add more responsive to the tooltip container max width, now it would not block off as much text as you shown in the screenshot

src/components/ColumnHeaderToolTip/LabelWithTooltip.tsx
tienhoah: this is changed

src/library/useTooltipPosition.js
tienhoah: it's TS now

src/components/ColumnHeaderToolTip/LabelWithTooltip.jsx
tienhoah: Yes, it's TS now

src/components/ColumnHeaderToolTip/ColumnHeaderToolTip.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+    useLayoutEffect(() => {
+      onMount?.()
+    }, [onMount])
+
+    return (
+      <TooltipPopup
+        ref={ref}
+        role="tooltip"
+        aria-labelledby="tooltip"
```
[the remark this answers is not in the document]
tienhoah: Good catch. `aria-labelledby="tooltip"` was referencing a hardcoded ID that doesn't exist in the DOM — effectively a no-op that would confuse screen readers. For role="tooltip", the element's text content already serves as the accessible name, so no additional labeling attribute is needed. Removed in the latest commit.

src/components/ColumnHeaderToolTip/LabelWithTooltip.jsx
tienhoah: Component is TSX now, and add null to type

gwan93 (approved):

alanjleonard (changes requested):
There should be only one tooltip visible at a time. If you click on one in the users table then the filter icon, the table should close and vise-versa. 

Image

alanjleonard:
> > Would it be possible to `align-items: flex-start` on the parent div so that there is less gap between the tooltip and the icon? Image
> > If the user has a really narrow screen the tooltip seems to get cut off. @alanjleonard do you have any thoughts on this? If none of our users use a narrow screen then this isn't an issue Image
> 
> 1. use align-items: flex-start now
> 2. I add more responsive to the tooltip container max width, now it would not block off as much text as you shown in the screenshot

I think that shorter text could help this as well. It doesn't need to be so long. That's another issue altogether. What you have works for now.

alanjleonard (approved):

PR #1601, opened 2026-06-12
feat [M585] Improve notification bell UX with shake animation

### LINKED ISSUE #1598: feat [M585] Improve notification bell UX with shake animation
## Problem Statement

The notification bell has low visibility when unread notifications exist. Users are either not noticing new notifications or not acting on them — resulting in a large backlog of unread notifications in the database. The current indicator is a small red dot that does not draw attention after the initial page load.

## Solution

Add a CSS shake/wiggle animation to the notification bell icon that plays automatically when unread notifications exist. The animation is attention-getting on first load and after new notifications arrive, but stops once the user opens the notification dropdown. Session state (via `sessionStorage`) ensures the animation does not repeat unnecessarily within a session, while still replaying when the user re-logs in without having acted on their notifications.

## Commits

1. **Add shake `@keyframes` animation and conditional CSS class to `BiggerIconBell`**
   Add a `@keyframes` shake animation (left-right rotation) to `Header.styles.js`. Add an `animated` prop/class variant to `BiggerIconBell` that applies the animation. Include a `prefers-reduced-motion` media query override that disables the animation for users who have that accessibility preference enabled. No behavior change yet — just CSS groundwork.

2. **Add 60-second polling interval to `useInitializeBellNotifications`**
   Introduce a `setInterval` (60s) that calls `updateNotifications()` periodically while the user is authenticated and online. Add proper cleanup (`clearInterval`) on unmount or when auth/online status changes. This ensures the notification count stays fresh for new-notification detection, independent of route changes.

3. **Add animation state tracking to `useInitializeBellNotifications`**
   Add the following state and logic to the hook:
   - `isAnimating: boolean` — controls whether the bell is currently animating
   - `animationLoopCount: number` — how many iterations to play (1 on app start, 3 on new notification / re-login)
   - `hasOpenedThisSession` — read/write from `sessionStorage` key (e.g. `notifications_opened`)
   - On initial load (500ms delay): if `notifications.length > 0` and `hasOpenedThisSession` is false, trigger animation with 1 loop
   - On polling result: if new notification count is higher than previous count, trigger animation with 3 loops
   - On re-login: if `isMermaidAuthenticated` transitions from false → true and `hasOpenedThisSession` is false and `notifications.length > 0`, trigger animation with 3 loops
   - Track previous notification count in a ref to detect increases
   - Expose `isAnimating`, `animationLoopCount`, and `markNotificationsOpened` from the hook and context

4. **Wire animation state to bell icon in Header (desktop and mobile)**
   In `Header.jsx`, consume `isAnimating` and `animationLoopCount` from `useBellNotifications()`. Pass them as props to `BiggerIconBell` to apply the animated class. Set `animation-iteration-count` via a CSS variable or inline style based on `animationLoopCount`. Apply to both desktop and mobile bell icon instances.

5. **Stop animation when user opens the notification dropdown**
   In `Header.jsx`, when the bell is clicked and the dropdown opens, call `markNotificationsOpened()`. This sets `hasOpenedThisSession` in `sessionStorage` and sets `isAnimating` to false, stopping the animation immediately.

6. **Update existing Header tests to cover notification indicator visibility**
   Extend `Header.test.jsx` to assert that the `NotificationIndicator` (red dot) is rendered when notifications are present and not rendered when the notifications array is empty. Use the existing test setup pattern in that file.

## Decision Document

- **Animation style**: CSS shake/wiggle (left-right rotation) — matches the bell metaphor
- **Accessibility**: `prefers-reduced-motion` media query disables animation entirely; red dot indicator remains unaffected
- **Animation loop counts**: 1 loop on app start; 3 loops on new notification arrival or re-login trigger
- **App start delay**: 500ms after component mount before triggering initial animation, to allow UI to settle
- **"Opened" definition**: User has clicked the bell and the dropdown has opened — does not require dismissing any notifications
- **Session persistence**: `sessionStorage` stores whether notifications were opened; cleared when the tab closes; survives logout/login within the same tab session
- **Unread count logic**: Same as the existing red dot — `notifications.length > 0` (notifications are either present or dismissed/deleted; no separate "viewed" status)
- **New notification detection**: 60-second polling interval comparing current count to previous count stored in a ref; if count increases, animation triggers
- **Polling interval**: 60s; can be tuned to 30s in future if no performance issues observed
- **Architecture**: Animation state folded into `useInitializeBellNotifications` and exposed via existing `BellNotificationContext` — no new context or global state added
- **Bell component**: Reuses existing `BiggerIconBell` styled component; no new icon or component added

## Testing Decisions

Good tests for this feature test external behavior (what the user sees), not implementation details (internal state variables or CSS class names).

- **What to test**: Whether the `NotificationIndicator` renders based on notification count — this is the existing behavior being extended
- **Module to update**: `Header.test.jsx` — add cases for indicator shown/hidden based on notifications array content
- **Prior art**: Existing Header tests use a mocked `useBellNotifications` context and render the Header component — follow the same pattern
- **What not to test**: CSS animation correctness, `sessionStorage` internals, or polling timing — these are implementation details

## Out of Scope

- No new test files — only existing `Header.test.jsx` is updated
- No changes to the notification dismiss/delete flow
- No changes to `BellNotificationDropDown` UI
- No "viewed" vs "dismissed" distinction — uses the same count logic as the red dot
- No real-time push or WebSocket — polling only
- No numeric unread count badge — binary indicator only (same as current red dot)
- No animation on the notification dropdown itself

## Further Notes

- The polling interval (`setInterval`) should be gated on `isMermaidAuthenticated && isAppOnline` to avoid unnecessary API calls
- The `sessionStorage` key should be namespaced to avoid collisions (e.g. `mermaid_notifications_opened`)
- Joe confirmed on Apr 8 2026 that the "loop 2-3 times per session" requirement was revised to "display once shortly after application starts" — the original looping behavior is replaced by the trigger-based approach documented here
- Animation iteration count differences between triggers (1 vs 3) can be implemented via a CSS custom property set inline, or via two utility classes — whichever is simpler given the styled-components setup
- **Known limitation**: if a new notification arrives after the user opens the bell (session flag set), a subsequent page refresh will not animate — the red dot still shows but the animation is suppressed for the rest of the session. A future improvement could store a "last seen count" in sessionStorage instead of a boolean flag, so the animation replays if the count increased since the bell was last opened

## Implementation Notes

Deviations from the original plan discovered during implementation and pre-PR review:

**Re-login animation race condition**
The original plan checked `notifications.length > 0` directly in the `isMermaidAuthenticated` effect. In practice, `isMermaidAuthenticated` flips to `true` before the notifications fetch resolves, so `notifications` is always `[]` at that point. Fixed with a `pendingReloginRef`: the auth effect sets the flag, and a separate notifications-watching effect consumes it once data arrives and triggers the 3-loop animation.

**`updateNotifications` wrapped in `useCallback`**
The polling `setInterval` captured a stale closure of `updateNotifications` because the function was recreated every render. Wrapped in `useCallback` with explicit deps so the interval always calls a fresh reference. Side effect: `isAppOnline` changes cause the interval to restart twice (once directly, once via the `logoutMermaid → handleHttpResponseError` dep chain) — harmless in practice.

**`aria-live` region added**
A visually-hidden `aria-live="polite"` span added to `Header.jsx` that announces "You have new notifications" to screen readers when `isAnimating` becomes true. Not in the original plan but added during pre-PR review for accessibility.

tienhoah:
## What changed
- Adds a CSS shake animation to the bell icon when unread notifications exist
- Polling every 60s detects new notifications and re-triggers the animation (3 loops)
- sessionStorage prevents the animation replaying unnecessarily within a session
- Animation stops immediately when the user opens the notification dropdown
- `prefers-reduced-motion` respected — animation disabled for users with that OS setting

## Test plan

**Setup:** ensure you have at least one unread notification on your account, _example: add or remove an user_

- [x] **Initial load (1 loop)** — close and reopen the tab (clears sessionStorage), log in. Bell shakes once after ~500ms.
- [x] **Session suppression** — open the bell dropdown, then hard-refresh. Bell does NOT animate.
- [x] **Polling detects new notification (3 loops)** — open the bell to mark it opened, then trigger a new notification (e.g. add/remove a user from a project). Within 60s, bell shakes 3 times.
- [x] **Re-login (3 loops)** — close and reopen the tab, log out, log back in. Bell shakes 3 times once notifications load.
- [x] **Re-login suppressed** — open the bell dropdown, log out, log back in. Bell does NOT animate.
- [x] **Reduce motion** — enable "Reduce Motion" in OS settings. Bell does NOT animate; red dot still shows.

Closes #1598

---Every PR---

---Transitional Changes---

---Post-Merge---

## Release Notes

* **New Features**
  * Bell icon now supports animated ringing when new notifications are detected, with reduced-motion support.
  * Added screen reader announcements (“You have new notifications.”) while the bell is animating.
* **Style**
  * Updated bell icon artwork and adjusted the notification indicator position.
* **Behavior**
  * Notification bell UI now updates via scheduled refresh after initial load.
* **Tests**
  * Added/updated tests to verify the notification indicator appears only when notifications exist.


src/App/useInitializeBellNotifications.js
```diff
@@ -58,6 +69,71 @@ export const useInitializeBellNotifications = ({
...
+    if (!wasAuthenticated && isMermaidAuthenticated) {
+      prevNotificationCountRef.current = null // reset so polling doesn't false-trigger after re-login
+      const hasOpenedThisSession =
+        sessionStorage.getItem(NOTIFICATIONS_OPENED_SESSION_KEY) === 'true'
+      if (!hasOpenedThisSession && notifications.length > 0) {
+        setAnimationLoopCount(3)
+        setIsAnimating(true)
+      }
+    }
+  }, [isMermaidAuthenticated]) // eslint-disable-line react-hooks/exhaustive-deps
```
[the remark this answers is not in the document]
tienhoah: Good catch. Fixed by adding a pendingReloginRef — the auth effect now sets the flag when isMermaidAuthenticated transitions false → true, and a separate notifications-watching effect consumes it once data arrives. This ensures the 3-loop animation fires after the fetch resolves rather than in the same render cycle as the auth change.

src/App/useInitializeBellNotifications.js
```diff
@@ -58,6 +69,71 @@ export const useInitializeBellNotifications = ({
...
+  useEffect(() => {
+    const prevCount = prevNotificationCountRef.current
+
+    if (prevCount !== null && notifications.length > prevCount) {
+      setAnimationLoopCount(3)
+      setIsAnimating(true)
+    }
+
+    prevNotificationCountRef.current = notifications.length
+  }, [notifications])
```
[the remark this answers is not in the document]
tienhoah: Thanks! This is intentional — the session flag only suppresses the initial-load animation on page refresh. If a new notification arrives via polling, we still want the bell to re-animate regardless, since it's a genuinely new event the user hasn't seen.

src/App/App.jsx
```diff
@@ -180,14 +180,21 @@ function App({ dexieCurrentUserInstance }) {
...
+    animationLoopCount,
+    markNotificationsOpened,
+    stopAnimation,
+  } = useInitializeBellNotifications({
+    apiBaseUrl,
+    getAccessToken,
+    isMermaidAuthenticated,
+    isAppOnline,
+    handleHttpResponseErrorWithLogoutAndSetServerNotReachableApplied,
+  })
```
gwan93, line 196: The values here are called and then just passed to BellNotificationProvider. BellNotificationProvider can call the hook itself to get these values so that they don't have to be drilled down.

```
// BellNotificationContext.jsx
export const BellNotificationProvider = ({ children, appConfig }) => {
  const { notifications, deleteNotification, deleteAllNotifications, isAnimating, animationLoopCount, markNotificationsOpened, stopAnimation } = useInitializeBellNotifications(appConfig);
  
  return (
    <BellNotificationContext.Provider value={{ notifications, deleteNotification, deleteAllNotifications, isAnimating, animationLoopCount, markNotificationsOpened, stopAnimation }}>
      {children}
    </BellNotificationContext.Provider>
  )
}
```

src/App/useInitializeBellNotifications.js
```diff
@@ -52,12 +64,95 @@ export const useInitializeBellNotifications = ({
...
 
+  // Poll for new notifications every 60s while authenticated and online
+  useEffect(() => {
+    if (!isMermaidAuthenticated || !isAppOnline) {
+      return undefined
+    }
+
+    const intervalId = setInterval(() => {
+      updateNotifications()
+    }, 60000)
```
gwan93: Could you define a constant at the top of the file for this this hardcoded 60000?

src/App/useInitializeBellNotifications.js
```diff
@@ -52,12 +64,95 @@ export const useInitializeBellNotifications = ({
...
+    hasTriggeredInitialAnimationRef.current = true
+    const hasOpenedThisSession = sessionStorage.getItem(NOTIFICATIONS_OPENED_SESSION_KEY) === 'true'
+    if (hasOpenedThisSession) {
+      return undefined
+    }
+
+    const timeoutId = setTimeout(() => {
+      setAnimationLoopCount(1)
+      setIsAnimating(true)
+    }, 500)
```
gwan93: Same with these? Something like `INITIAL_ANIMATION_LOOP_COUNT` and `ANIMATION_TRIGGER_DELAY_MS`

src/App/useInitializeBellNotifications.js
```diff
@@ -52,12 +64,95 @@ export const useInitializeBellNotifications = ({
...
+
+    return () => clearTimeout(timeoutId)
+  }, [notifications])
+
+  // Trigger animation when polling detects new notifications
+  useEffect(() => {
+    const prevCount = prevNotificationCountRef.current
+
+    if (prevCount !== null && notifications.length > prevCount) {
+      setAnimationLoopCount(3)
```
gwan93: Here as well

gwan93 (changes requested):
Few minor changes

alanjleonard (changes requested):
Looks good. I think you can just loop it once, not 3 times.

src/App/App.jsx
tienhoah: I'm following the established pattern of other hooks in App.jsx — it serves as the initialization orchestrator, and the bell hook fits naturally there. Down the road, we could refactor the initialization pipeline to use providers (a large architectural refactor), but that's a separate effort.

src/App/useInitializeBellNotifications.js
tienhoah: Updated

src/App/useInitializeBellNotifications.js
tienhoah: Updated

src/App/useInitializeBellNotifications.js
tienhoah: Updated

tienhoah:
> Looks good. I think you can just loop it once, not 3 times.

I change to loop one time now, its more simple this way. Please check again!

src/App/App.jsx
gwan93: The other hooks in App.jsx all use the values to derive/calculate something else (with the exception of one value). The ones used for the bell notification are all directly drilled through to the provider so they don't need to be here

I agree we can do a separate architectural refactor later if there is ever time

gwan93 (approved):

alanjleonard (approved):

PR #1607, opened 2026-06-16
hot fix: [M1511] tooltip not closing when clicking on table inputs

tienhoah:
## What changed

- Tooltips on observation table column headers (Fish Belt, Benthic, etc.) would not close when clicking anywhere in the table body — inputs, autocompletes, or empty header space
- Root cause: Downshift (the autocomplete library) registers a capture-phase `mousedown` listener on `document` and calls `stopImmediatePropagation()`, which blocked `ClickAwayListener` from ever firing
- Fix: replace `ClickAwayListener` with a transparent `position: fixed` backdrop rendered at z-index 1499 (just below MUI Tooltip at 1500) — any click outside the tooltip lands directly on the backdrop, bypassing event propagation entirely

## Test plan

- [ ] Open a tooltip on any observation table column header (Fish Belt, Benthic LIT, Benthic PIT, etc.)
- [ ] Click on an autocomplete input in the table body — tooltip should close
- [ ] Click on a count/size input in the table body — tooltip should close
- [ ] Click in the empty space of a wide column header — tooltip should close
- [ ] Click the info icon again — tooltip should toggle closed
- [ ] Open two tooltips in sequence — only one should be visible at a time (groupRef behavior unchanged)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tooltip dismissal behavior to be more responsive and reliable when clicking outside the tooltip area.

### DIFF
--- a/src/components/ColumnHeaderToolTip/LabelWithTooltip.tsx
+++ b/src/components/ColumnHeaderToolTip/LabelWithTooltip.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react'
+import React, { useState, useEffect, useRef } from 'react'
 import Tooltip from '@mui/material/Tooltip'
 import { LabelContainer } from '../generic/form'
 import { IconInfo } from '../icons'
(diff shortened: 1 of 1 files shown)


gwan93 (commented):
When I have the Fish name tooltip open and then I click on the cell to the right of it (still in the header), in this case clicking on the "Size (cm)" text or anywhere within that cell, the fish name tooltip still persists.

Everything else seems to work though

Image

gwan93 (approved):

PR #1618, opened 2026-06-23
fix [1997]

gwan93:
update form size behavior to clear observation size values when the user changes the macroinvertebrate size bin with a modal confirmation. ensure observation table is always enabled, remove warning text, remove requirement for macroinvertebrate size field

---Every PR---

---Transitional Changes---

---Post-Merge---

## Summary by CodeRabbit

* **New Features**
  * Added a confirmation dialog for clearing/applying belt invert size values when changing the size bin while size data already exists.

* **Improvements**
  * Improved one-decimal formatting and sanitization for belt invert size input, including clearer focused-vs-finalized input behavior.

* **Bug Fixes**
  * Adjusted when the “add row” button becomes disabled to rely only on loading error state.

* **Tests**
  * Updated belt invert validation display integration tests to match the refined validation visibility behavior.

* **Documentation**
  * Updated the size-bin tooltip text to reflect the active bin scheme.

### DIFF
--- a/src/components/pages/submittedRecordPages/SubmittedBeltInvert/SubmittedBeltInvertObservationTable.jsx
+++ b/src/components/pages/submittedRecordPages/SubmittedBeltInvert/SubmittedBeltInvertObservationTable.jsx
@@ -54,11 +54,13 @@ const SubmittedBeltInvertObservationTable = ({
     return roundToOneDecimal(observationDensities.get(item.id))
   }
 
+  const hasSizeData = obs_belt_inverts.some((item) => item.size)
+
   const observationBeltInverts = obs_belt_inverts.map((item, index) => (
     <Tr key={item.id}>
       <Td $align="center">{index + 1}</Td>
       <Td $align="left">{getInvertName(item.invert_attribute)}</Td>
-      <Td $align="left">{item.size}</Td>
+      {hasSizeData && <Td $align="left">{item.size}</Td>}
       <Td $align="right">{item.count}</Td>
       <Td $align="right">{item.notes}</Td>
       <Td $align="right">{getInvertDensity(item)}</Td>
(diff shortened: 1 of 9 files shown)


alanjleonard (changes requested):
In the helper text you can remove "Choose 1 cm if macroinvertebrate size does not use bins." since it's not required.
Otherwise approved.

alanjleonard (changes requested):
We don't need to show the `size_bin_required` validation in the table. This is also blocking the user from submitting the record.

Image

gwan93:
> We don't need to show the `size_bin_required` validation in the table. This is also blocking the user from submitting the record.
> 
> Image

Good catch on this! 

The frontend simply displays the validation errors received from the backend so the backend will need to be updated to make the size bin not required. No work will be necessary on the frontend to reflect the backend change. I'll follow up with Kim regarding this

alanjleonard (approved):

alanjleonard (approved):

src/components/pages/submittedRecordPages/SubmittedBeltInvert/SubmittedBeltInvertObservationTable.jsx
```diff
@@ -54,11 +54,13 @@ const SubmittedBeltInvertObservationTable = ({
...
   }
 
+  const hasSizeData = obs_belt_inverts.some((item) => item.size)
+
   const observationBeltInverts = obs_belt_inverts.map((item, index) => (
     <Tr key={item.id}>
       <Td $align="center">{index + 1}</Td>
       <Td $align="left">{getInvertName(item.invert_attribute)}</Td>
-      <Td $align="left">{item.size}</Td>
+      {hasSizeData && <Td $align="left">{item.size}</Td>}
```
tienhoah: `item.size` is rendered raw here. The edit view formats size to one decimal in its resting state — is there an expected display format for size in the submitted view? If so, can you make it consistent?

src/locales/en/translation.json
```diff
@@ -534,7 +534,7 @@
     "macroinvertebrate_species": "Macroinvertebrate species",
     "must_select_size_bin_warning": "You must select a macroinvertebrate size bin before adding any observations",
```
tienhoah: This is not used

tienhoah (changes requested):
Almost there! 2 minor requests

tienhoah (approved):

PR #1622, opened 2026-06-24
fix [2013] change num of decimals displayed

gwan93:
Ref: https://trello.com/c/Uqc7cE4o/2013-macroinvertrebrate-summary-table-needs-to-use-the-same-number-of-decimal-places

Note: the summary looks a little different because another PR with the format changes is waiting to be reviewed

image

---Every PR---

---Transitional Changes---

---Post-Merge---


alanjleonard (approved):

src/library/macroinvertebrates/useBeltInvertDensityMetrics.ts
```diff
@@ -52,8 +52,8 @@ interface BeltInvertObservationMetrics {
   observationDensities: Map<string, number>
 }
 
-export const formatDensityToTwoDecimals = (value: number) =>
-  (Math.round(Number(value) * 100) / 100).toFixed(2)
+export const formatDensityToOneDecimal = (value: number) =>
+  (Math.round(Number(value) * 100) / 100).toFixed(1)
```
JBurkinshaw: When this rounded to two decimals the `Math.round(value * 100) / 100` matched the `.toFixed(2)` and was fine. Now that we output one decimal, it double-rounds: it rounds to two decimals first, then `.toFixed(1)` rounds again. That shifts some values, e.g. `1.249` → `1.3` (should be `1.2`), `1.45` → `1.4` (should be `1.5`), `2.349` → `2.4` (should be `2.3`).

We already have `roundToOneDecimal` in `library/numbers`, which does the correct single-step round, and both files in this PR already import it (used a few lines away).

I suggest deleting `formatDensityToOneDecimal` and calling `roundToOneDecimal` at the four call sites instead. That fixes the rounding and drops a near-duplicate helper. (Values here are typed `number`, so the missing `Number()` wrapper in `roundToOneDecimal` isn't an issue.)

Note that `roundToOneDecimal` has no unit test, so this consolidation doesn't add coverage. Worth a small `roundToOneDecimal.test.js` pinning the boundary cases (`1.249 → "1.2"`, `1.45 → "1.5"`) so the display precision is protected going forward. Something along these lines:
```js
import { roundToOneDecimal } from './roundToOneDecimal'

describe('roundToOneDecimal', () => {
  it.each([
    [1.249, '1.2'],
    [1.45, '1.5'],
    [2.349, '2.3'],
    [0, '0.0'],
  ])('rounds %f to %s', (input, expected) => {
    expect(roundToOneDecimal(input)).toBe(expected)
  })
})
```

JBurkinshaw (commented):
Correct intent, minimal and well-scoped. The one thing I'd change before merge is dropping the redundant helper in favour of `roundToOneDecimal` (see comment in code) - it fixes a rounding discrepancy and removes duplication, with essentially no added risk since the helper is already imported in both files

JBurkinshaw (approved):

PR #1626, opened 2026-06-26
[M2033] GoI equal-split weighting

saanobhaai:
## Summary by CodeRabbit

* **Bug Fixes**
  * Updated density calculations so family, order, and class observations are split evenly across distinct applicable group-of-interest values.
  * Ensured zero-count observations no longer produce density results, and very small rounded values are excluded as expected.

* **Tests**
  * Added coverage for direct and shared attribution behavior, total density calculation, and rounding edge cases.

### DIFF
--- a/src/library/macroinvertebrates/calculateGroupOfInterestDensity.test.ts
+++ b/src/library/macroinvertebrates/calculateGroupOfInterestDensity.test.ts
@@ -108,12 +108,25 @@ describe('calculateDensityByGroupOfInterest', () => {
   })
 
   describe('ghost-zero filter', () => {
-    it('omits GoIs whose density rounds to 0.00', () => {
-      // count=0 observation should produce no GoI entries
+    it('skips zero-count observations', () => {
       const obs = [{ count: 0, invert_attribute: 'family-1' }]
       const { densityByGoi } = calculateDensityByGroupOfInterest(obs, allAttributes, LEN, WIDTH, { goiChoices })
 
       expect(Object.keys(densityByGoi)).toHaveLength(0)
+    })
+
+    it('omits GoIs whose density rounds to 0.00 and keeps those above', () => {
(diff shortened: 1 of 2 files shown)


saanobhaai:
[client-side GoI equal-split weighting](https://trello.com/c/o2ZGmv8k/2033-client-side-goi-equal-split-weighting)

src/library/macroinvertebrates/calculateGroupOfInterestDensity.test.ts
```diff
@@ -0,0 +1,119 @@
...
+  describe('ghost-zero filter', () => {
+    it('omits GoIs whose density rounds to 0.00', () => {
+      // count=0 observation should produce no GoI entries
+      const obs = [{ count: 0, invert_attribute: 'family-1' }]
+      const { densityByGoi } = calculateDensityByGroupOfInterest(obs, allAttributes, LEN, WIDTH, { goiChoices })
+
+      expect(Object.keys(densityByGoi)).toHaveLength(0)
+    })
+  })
+})
```
JBurkinshaw, line 132: This one come courtesy of claude:

This test passes, but not for the reason the name implies. A `count: 0` observation hits the `count <= 0` guard in the main loop and returns before any GoI attribution happens, so `goiCountTotals` ends up empty and `densityByGoi` is empty simply because was ever added - the `rounded > 0` ghost-zero filter never runs. So this is really testing the zero-count short-circuit, not the rounding filter it's named after.

To actually cover the filter, use a small *positive* count whose density rounds to `0.00` (i.e. `< 0.005` ind/ha) and assert that GoI is dropped while a larger one survives - e.g. a tiny count against a tagged genus with a large area. Could also keep this one and rename it to something like "skips zero-count observations" so both paths are covered.

JBurkinshaw (approved):

PR #1634, opened 2026-06-29
fix [2011] align observation table headers and values

gwan93:
Ref: https://trello.com/c/Csdl9rCq/2011-macroinvertrebrate-observation-table-data-alignment

Align table headers associated with numeric values to the right
Align table values with type of number to the right

Before:
image
image

After:
image

image

---Every PR---

---Transitional Changes---

---Post-Merge---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved alignment for belt invert observation tables so numeric values and the size column render right-aligned consistently.
  * Refined observation table headers and density units for clearer label/unit presentation.
  * Updated size formatting/visibility logic to handle truly non-empty values correctly (including `0`).

* **New Features**
  * Added shared “non-empty value” handling to the size selector behavior.

* **Tests**
  * Added an offline integration test validating belt invert size formatting after updates.
  * Added unit tests for `hasNonEmptyValue`.

* **Style**
  * Updated table cell spacing and alignment styling to match the shared theme.


src/components/pages/collectRecordFormPages/BeltInvertForm/BeltInvertObservationTable.tsx
```diff
@@ -310,8 +312,9 @@ const BeltInvertObservationRow = ({
         </InputAutocompleteContainer>
       </Td>
       {sizeBinSelectedLabel && <Td $align="right">{sizeInput}</Td>}
-      <Td $align="right">
+      <Td>
         <InputNumberNumericCharactersOnly
+          $textAlign="right"
```
JBurkinshaw, line 318: The size cell keeps `$align="right"` on the `<Td>`, but the count cell drops it and relies only on `$textAlign="right"` on the input. Both inputs are `width: 100%` (from `inputTextareaSelectStyles`), so the `<Td>`'s `text-align` likely has no visible effect on a full-width input - the input's own `$textAlign` is what positions the value. That makes the `$align="right"` on the size `<Td>` dead code. Can you drop it so both cells are styled the same way the count cell is?
e.g.
```js
{sizeBinSelectedLabel && <Td>{sizeInput}</Td>}`
```

src/library/hasNonEmptyValue.js
```diff
@@ -0,0 +1,3 @@
+export const hasNonEmptyValue = (value) => {
```
JBurkinshaw: This is now a shared library utility, but it's a plain `.js` file with no test. The PR checklist says to migrate touched/new JS to TypeScript where in scope. This file is brand new it could have started as `.ts`. A short test would be good for one subtlety that matters here - that `0` and `false` are treated as non-empty while `''`, `null`, and `undefined` are empty.
Rename to `hasNonEmptyValue.ts` and add a short test covering `0`, `''`, `null`, `undefined`, and a normal value.

src/components/pages/submittedRecordPages/SubmittedBeltInvert/SubmittedBeltInvertObservationTable.jsx
```diff
@@ -61,13 +62,17 @@ const SubmittedBeltInvertObservationTable = ({
...
+  const hasSizeData = obs_belt_inverts.some((item) => hasNonEmptyValue(item.size))
 
   const observationBeltInverts = obs_belt_inverts.map((item, index) => (
     <Tr key={item.id}>
       <Td $align="center">{index + 1}</Td>
       <Td $align="left">{getInvertName(item.invert_attribute)}</Td>
-      {hasSizeData && <Td $align="left">{item.size ? roundToOneDecimal(item.size) : ''}</Td>}
+      {hasSizeData && (
+        <Td $align="right">
+          {hasNonEmptyValue(item.size) ? roundToOneDecimal(item.size) : ''}
```
JBurkinshaw, line 73: The behaviour change where a size of `0` now renders as `0.0` (and the size column stays visible when every observation is `0`) is exactly the kind of thing that could cause a regression. There's no test asserting it. If it's easy to add, a test that renders a row with `size: 0` and checks for `0.0` could be worthwhile.

JBurkinshaw (changes requested):

JBurkinshaw:
Nice work. A clean, well-scoped fix that correctly finds the root cause (the flex wrapper overriding `text-align`) rather than working around it. The `0` handling is also a good catch

alanjleonard (changes requested):
The density column should have the same padding as the other columns. Right now it has none. Otherwise approved.

alanjleonard (changes requested):
The size dropdown needs to be aligned right as well when you select a size bin.

Image

alanjleonard (approved):

JBurkinshaw (approved):

PR #1636, opened 2026-06-29
fix [m2014] update notes

gwan93:
Ref:https://trello.com/c/t6RFYjS3/2014-macroinvertrebrate-observations-table-style-updates-to-notes-column-and-keybaord-shortcuts

---Every PR---

---Transitional Changes---

---Post-Merge---


alanjleonard (changes requested):
- `enter` should still make a blank row when any cell is selected. This is the existing behaviour on other observation tables.
- If there's notes in the cell, show the notes when focused not "Click or press space to add notes"
- Textarea should not start with a space (from pressing space to open it)

alanjleonard (approved):

src/components/pages/collectRecordFormPages/BeltInvertForm/EditNotesModal.module.scss
```diff
@@ -20,6 +20,7 @@
 
   &:hover {
     background: theme.$tableRowEven;
+    border: theme.$border-medium solid theme.$primary;
```
JBurkinshaw: The ticket asked for "the blue outline that other editable cells have." Every other editable cell uses `outline` (see `src/style/forms.module.scss:19-22`, and this cell's own `:focus` rule right below). Here the hover state uses `border` instead. Because the cell has no border by default, adding a 2px border only on hover grows the box by 2px on each side, so the cell contents visibly shift/jump every time the pointer enters. Can you switch hover to match focus - `outline: theme.$border-medium solid theme.$primary; outline-offset: -2px;`?

Something like this:
```jsx
&:hover {
  background: theme.$tableRowEven;
  outline: theme.$border-medium solid theme.$primary;
  outline-offset: -2px;
}
```

src/locales/en/translation.json
```diff
@@ -529,8 +529,9 @@
   "loading_indicator": "Loading indicator",
   "loading_item": "Loading {{itemName}}",
   "macroinvertebrate_observations": {
-    "add_notes": "Add notes...",
+    "add_notes": "Click to add notes...",
     "add_macroinvertebrate": "Add macroinvertebrate species",
+    "click_or_press_space_to_add_notes": "Click or press space to add notes",
```
JBurkinshaw: The ticket shows "…" on both strings and `add_notes` keeps its dots, but the new token drops them. Add "..." for consistency

JBurkinshaw (changes requested):

JBurkinshaw (approved):

PR #1639, opened 2026-06-30
fix [m2007] fix dropdown padding and styling

gwan93:
Ref: https://trello.com/c/NA7EB6Ss/2007-add-padding-to-the-genus-dropdown-when-proposing-a-new-species

Fix the padding and styling on the dropdowns

After fixes:
image
image

---Every PR---

---Transitional Changes---

---Post-Merge---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the autocomplete “no results” experience so the message and action now render consistently within the dropdown.
  * Kept the “contact us” link and no-results text visible while simplifying the layout.
  * Updated dropdown item styling to better support text, links, and buttons across autocomplete results.


alanjleonard (approved):

src/components/generic/InputAutocomplete/InputAutocomplete.jsx
```diff
@@ -30,13 +30,6 @@ const AutoCompleteResultsWrapper = styled.div`
       padding: ${theme.spacing.buttonPadding};
     }
   }
-
-  button {
-    width: 100%;
-    border: none;
-    text-align: start;
-    padding: ${theme.spacing.small};
-  }
```
JBurkinshaw, line 39: I think the entire `& > div { ... }` block is dead now.  The only direct children of this wrapper are `AutoCompleteInput` (renders an `<input>`, since `Input` is `styled.input`), `HelperText` (a `<span>`), and `Menu` (a `<ul>`). Nothing renders a direct `<div>` child any more - that block existed solely to position the old no-results div you just removed, and the menu positions itself via its own `position: absolute; top: 4rem` in `InputAutocomplete.styles.js`. Can you delete the whole block rather than just the paragraph rule?

src/components/generic/InputAutocomplete/InputAutocomplete.jsx
```diff
@@ -178,19 +171,12 @@ const InputAutocomplete = ({
                 'data-testid': menuTestId,
               })}
             >
-              {isMenuOpen && getMenuContents(downshiftObject)}
+              {isMenuOpen && menuItems.length > 0 && getMenuContents(downshiftObject)}
+              {isMenuOpen && !menuItems.length && noResultsText && (
+                <Item data-testid="noResult">{noResultsText}</Item>
+              )}
+              {isMenuOpen && !menuItems.length && noResultsAction && <Item>{noResultsAction}</Item>}
```
JBurkinshaw, line 178: Small thing. `getMenuProps` marks this `<ul>` as a listbox, and the real options get an `option` role via `getItemProps`. These two no-results `<Item>`s don't, so they're plain `<li>`s sitting inside a listbox - which screen readers can misread as empty/extra options. They used to live outside the menu, so this is a slight step back on that front. If you feel like tidying it, rendering them outside the `getMenuProps` element (or wrapping in `role="presentation"`) should fix it.

JBurkinshaw (commented):
Couple of minor comments in this one

PR #1643, opened 2026-07-02
fix [m1958] conditionally render header in summary metrics

gwan93:
Ref: https://trello.com/c/d2Wkqk29/1958-idea-macroinvertebrate-submitted-record-view-calculated-metrics-phase-41

Conditionally render the "Density by group of interest (ind/ha)" text only when the record has observations

Collect:
image

Submitted:
image

---Every PR---

---Transitional Changes---

---Post-Merge---

## Summary by CodeRabbit

* **Bug Fixes**
  * The “density by group of interest” section now only renders when density data is available, avoiding empty table headers in belt invert observation views.
  * This behavior is consistent across both record entry and submitted record pages.
* **Refactor**
  * Macroinvertebrate density/abundance summary rendering was consolidated into a shared table component and reused across belt invert pages.


alanjleonard (approved):

src/components/pages/collectRecordFormPages/BeltInvertForm/BeltInvertObservationTable.tsx
```diff
@@ -616,13 +616,15 @@ const BeltInvertObservationTable = ({
...
-          </thead>
+          {Object.entries(densityByGoi).length > 0 && (
+            <thead>
+              <Tr>
+                <Th colSpan={2}>
+                  {t('macroinvertebrate_observations.density_by_group_of_interest_units')}
+                </Th>
+              </Tr>
+            </thead>
+          )}
```
JBurkinshaw: The two `MacroinvertebrateObservationsSummaryStats` blocks are now essentially identical. You've applied the fix in both places, but the same block is being maintained twice. It'd be worth extracting a small shared `<MacroinvertebrateSummaryStats densityByGoi={...} totalDensity={...} abundance={...} />` so the next change to this table only happens once. The two `roundToOneDecimal`/`toFixed(1)` conventions would come along as part of that.

JBurkinshaw (changes requested):

JBurkinshaw (approved):

PR #1650, opened 2026-07-06
fix [m2039] limit input to one number after decimal

gwan93:
Ref: https://trello.com/c/82Jyn6uD/2039-bug-invert-size-bin-input-behavior

Limit how many numbers can be entered after the decimal place to just 1 number

image

---Every PR---

---Transitional Changes---

---Post-Merge---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of “50+” size entries so inputs are cleaned up, limited to one decimal place, and saved only when valid.
  * Fixed display behavior for invalid or incomplete numeric entries in the size input.
  * Corrected rounding/truncation for “>50” values, ensuring entries like `52.3456` display as `52.3`.
* **Tests**
  * Added broader test coverage for “50+” size input behavior across multiple size-bin options.


src/components/pages/collectRecordFormPages/ObservationSizeSelect.jsx
```diff
@@ -41,21 +54,25 @@ const ObservationSizeSelect = ({
   const handleSelectOnChange = (event) => {
     const selectedValue = event.target.value
     const isSelectedValue50 = selectedValue === '50'
-    const valueToSubmit = isSelectedValue50 ? plus50Value : selectedValue
+    const valueToSubmit = isSelectedValue50
```
JBurkinshaw: The select's first option is `<option value=""> </option>` (the clear/placeholder). When a user picks it, `selectedValue` is `""`, and `sanitizeNumericDecimalInput("")` returns `0`. So instead of clearing the size, we now store `0`. Previously this passed `""` straight through. The select options come from a fixed bin list, they never contain stray characters or multiple decimalsm so running them through the sanitizer buys nothing and introduces this regression. Can you leave it untouched?

src/components/pages/collectRecordFormPages/ObservationSizeSelect.jsx
```diff
@@ -41,21 +54,25 @@ const ObservationSizeSelect = ({
...
-    setPlus50Value(event.target.value)
+    const sanitizedValue = sanitizeNumericDecimalInput(event.target.value)
+    setPlus50Value(sanitizedValue)
   }
 
   const handlePlus50OnBlur = (event) => {
-    const eventValue = Number(event.target.value)
-    const validPlus50Value = Number.isFinite(eventValue) && eventValue >= 50 ? eventValue : ''
+    const sanitizedValue = sanitizeNumericDecimalInput(event.target.value)
+    const isValidPlus50Value = sanitizedValue >= 50
```
JBurkinshaw, line 70: In the invalid branch you set the local state to `50` but submit `''`. Submitting `''` makes the parent re-render with `value=''`, which triggers the effect (`!hasNonEmptyValue`) that sets `plus50Value` to `50` and hides the input anyway. This means the `setPlus50Value(50)` here is dead and a bit misleading (it reads as "reset the box to 50" but the box disappears). I'd drop it and do something like `onValueEntered(isValidPlus50Value ? sanitizedValue : '')` to keep the single source of truth in the effect.

JBurkinshaw (changes requested):

alanjleonard (approved):

JBurkinshaw (approved):

PR #1652, opened 2026-07-07
[M2040] allow 0 for number_of_solutions_supported_by

saanobhaai:
## Summary by CodeRabbit

* **New Features**
  * Integer input fields can now display a “required” state when needed (including support for alternate compact/no-row layouts).
* **Bug Fixes**
  * Updated the finance solution form’s “number of solutions supported” field to use the standardized integer input, improving numeric entry and validation behavior across devices.
  * Default values for “number of solutions supported” now load more reliably (including showing `0` instead of ending up blank/unexpected).


saanobhaai:
['Number of solutions supported by' field does not accept zero as a value in 'Technical assistance facility', 'Conservation trust fund' and 'Financial facility' in GFCR page](https://trello.com/c/1tVnPH6Z/2040-number-of-solutions-supported-by-field-does-not-accept-zero-as-a-value-in-technical-assistance-facility-conservation-trust-fund)

saanobhaai:
['Number of solutions supported by' field does not accept zero as a value in 'Technical assistance facility', 'Conservation trust fund' and 'Financial facility' in GFCR page](https://trello.com/c/1tVnPH6Z/2040-number-of-solutions-supported-by-field-does-not-accept-zero-as-a-value-in-technical-assistance-facility-conservation-trust-fund)

saanobhaai:
['Number of solutions supported by' field does not accept zero as a value in 'Technical assistance facility', 'Conservation trust fund' and 'Financial facility' in GFCR page](https://trello.com/c/1tVnPH6Z/2040-number-of-solutions-supported-by-field-does-not-accept-zero-as-a-value-in-technical-assistance-facility-conservation-trust-fund)

saanobhaai:
@coderabbitai review

src/components/pages/gfcrPages/GfcrIndicatorSetForm/modals/FinanceSolutionModal.jsx
```diff
@@ -29,6 +29,10 @@ import { getToastArguments } from '../../../../../library/getToastArguments'
 import { useHttpResponseErrorHandler } from '../../../../../App/HttpResponseErrorHandlerContext'
 import InputNoRowWithLabelAndValidation from '../../../../mermaidInputs/InputNoRowWithLabelAndValidation'
 import InputNoRowSelectWithLabelAndValidation from '../../../../mermaidInputs/InputNoRowSelectWithLabelAndValidation'
+import {
+  formikHandleGfcrNumberInputChange,
+  formikHandleIntegerInputOnBlur,
+} from '../../../../../library/formik/formikHandleInputTypes'
```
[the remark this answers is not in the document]
saanobhaai: Verified — leaving as-is. GfcrIntegerInputField was the original implementation here, but it hardcodes InputWithLabelAndValidation, which wraps in <InputRow> (generic/form.js:86-103) — a grid layout with a colored left border and hover highlight. That's neutralized in the F1–F7 forms via a .inputRow CSS override (subPages.styles.js:10-27), but FinanceSolutionModal has no such override, and every sibling field here (Type, Facility/solution name, sector, taf_name, etc.) uses the plain InputNoRowWithLabelAndValidation with no grid/border. Using GfcrIntegerInputField caused this field to render with a mismatched blue left-border/grid layout that didn't match anything else in the modal — confirmed visually, not just in code.

What's duplicated here is two thin wrapper closures plus three JSX attributes (type/inputMode/pattern). The actual logic — digit-stripping on change, reset-to-initial on blur — isn't duplicated; it's imported directly from the same shared module (formikHandleGfcrNumberInputChange/formikHandleIntegerInputOnBlur) that GfcrIntegerInputField itself calls. That's the right level of reuse for this spot; pulling in the whole presentational component is what caused the layout bug. Keeping the current implementation.

JBurkinshaw (approved):
Looks good to me

src/components/pages/gfcrPages/GfcrIndicatorSetForm/GfcrIntegerInputField.jsx
```diff
@@ -40,6 +41,7 @@ const GfcrIntegerInputField = ({
       pattern="[0-9]*"
       helperText={helperText}
       showHelperText={displayHelp}
+      required={required}
```
JBurkinshaw, line 47: The `required` prop looks out of scope for this PR.

The prop is wired up correctly - it forwards through to `InputWithLabelAndValidation`, which does render a required indicator from it. But no caller passes it into `GfcrIntegerInputField` (not the modal this PR fixes, nor the F3/F5/F6/F7 forms), so it doesn't change anything as part of the zero fix. Is it out of scope here? Happy for it to stay if there's a good reason, otherwise might be cleaner to drop it from this PR.

src/components/pages/gfcrPages/GfcrIndicatorSetForm/modals/FinanceSolutionModal.jsx
```diff
@@ -195,19 +199,26 @@ const FinanceSolutionModal = ({
...
+  }
+
+  const handleNumberOfSolutionsBlur = (event) => {
+    formikHandleIntegerInputOnBlur({
+      formik,
+      event,
+      fieldName: 'number_of_solutions_supported_by',
+    })
+  }
+
```
JBurkinshaw: Minor, but worth doing here. This part re-implements logic that `GfcrIntegerInputField` already has. Both use the same change/blur helpers with the same settings, so the behavior is identical - it's just been wired up again instead of reused.

It likely wasn't reused because `GfcrIntegerInputField` is hardwired to the row layout, but the modal needs the stacked one. If we add a `noRow` option to `GfcrIntegerInputField` so it can render the stacked variant, the modal can drop these two handlers (and the helper imports) and just use the component:

`<GfcrIntegerInputField id="number_of_solutions_supported_by" label={...} formik={formik} required noRow />`

That leaves one copy of the integer-input logic instead of two, which matters because two copies can drift.

JBurkinshaw (changes requested):

JBurkinshaw (approved):
I made a few changes to Kim's code based on my own feedback

src/components/pages/gfcrPages/GfcrIndicatorSetForm/GfcrIntegerInputField.jsx
saanobhaai: Again not my primary domain, but I think it is used, by the modal. The old manual formik validation (Number(values.number_of_solutions_supported_by) <= 0 → custom "Required" error) was deleted, and the field's required={true} (previously passed straight to InputNoRowWithLabelAndValidation) was carried over to the new GfcrIntegerInputField wrapper instead.

PR #1662, opened 2026-07-10
[2047] allow non-last admin to remove self from project

saanobhaai:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved user removal handling so the dialog stays open on failure and preserves error feedback.
  - Removing yourself from a project now redirects to the projects list.
  - Role changes now refresh the current user immediately for updated access control.
  - Prevented removing or demoting the sole project administrator, with updated tooltip guidance.
  - Added a clearer error message when user removal is unsuccessful.

### DIFF
--- a/src/components/RemoveUserModal/RemoveUserModal.jsx
+++ b/src/components/RemoveUserModal/RemoveUserModal.jsx
@@ -101,7 +101,9 @@ const RemoveUserModal = ({
   const footerContentPageTwo = (
     <RightFooter>
       {cancelButton}
-      <ButtonCaution onClick={handleOnSubmit}>{t('users.remove_user')}</ButtonCaution>
+      <ButtonCaution onClick={handleOnSubmit} disabled={isLoading}>
+        {t('users.remove_user')}
+      </ButtonCaution>
     </RightFooter>
   )
 
--- a/src/components/RemoveUserModal/RemoveUserModal.jsx
+++ b/src/components/RemoveUserModal/RemoveUserModal.jsx
@@ -38,9 +38,14 @@ const RemoveUserModal = ({
   }
 
   const handleOnSubmit = () => {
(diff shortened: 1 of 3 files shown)


saanobhaai:
[allow non-last admin to remove self from project](https://trello.com/c/e0KuaDJn/2047-allow-non-last-admin-to-remove-self-from-project)

src/components/RemoveUserModal/RemoveUserModal.jsx
JBurkinshaw: Now that `removeUserProfile` returns the real promise chain (this looks like a good change), the modal correctly stays open for the whole request instead of closing almost immediately like before. The "Remove user" button has no `disabled` property, so while we're waiting for the promise to resolve you could click it a few times and fire off multiple `removeUser` calls.

Call path: `handleOnSubmit()` → `onSubmit()` (= `Users.removeUserProfile()`) → `removeUser(...)`, and the modal only calls `closeModal()` once that resolves. Previously `removeUserProfile` returned `Promise.resolve()` synchronously so the modal closed almost immediately, before a second click was possible... now it's reachable for the full round-trip.

The button should be disabled while waiting for the promise. We're already passing `isLoading={isTableUpdating}` into the modal, it's just unused on this button, so a fix would be a one-liner:
> ```jsx
> <ButtonCaution onClick={handleOnSubmit} disabled={isLoading}>{t('users.remove_user')}</ButtonCaution>

JBurkinshaw (changes requested):
This looks good. Just one small change required

src/components/RemoveUserModal/RemoveUserModal.jsx
saanobhaai: @JBurkinshaw addressed

JBurkinshaw (approved):

PR #1665, opened 2026-07-15
fix: [M2048] Key Collect status colour on stable value

JBurkinshaw:
### Summary
On the Collect page each sample unit row has a coloured left-border indicating its status. The colour lookup was keyed on the English status label (`RECORD_STATUS_LABELS`), but the rendered status is now translated. In any non-English language the lookup missed and the border colours vanished.

### Changes

- `CollectRecordsMixin.js`: new `#getStatusValue` normalises the raw validation status to a stable key (`error`/`warning`/`ok`/`stale`, folding `undefined`/offline records into `stale`); `#getStatusLabel` now derives from it, and `uiLabels.statusValue` is exposed.
- `Collect.jsx`: the row border is keyed on the stable `statusValue` rather than the translated label. The `RECORD_STATUS_COLORS` map (still sourced from `theme.color`) now lives here and reaches the row through a `--record-status-color` CSS custom property.
- `Collect.module.scss` (new): the border structure moves out of the styled-component into an SCSS module that consumes the CSS custom property.
- `Collect.styles.js`: removed; its only export (`TrCollectRecordStatus`) is replaced by the SCSS module applied to `<Tr>`.
- `collectConstants.js` → `collectConstants.ts`: migrated to TypeScript; removed the now-unused `RECORD_STATUS_LABELS`.
- Added `collectRecordStatusValue.databaseSwitchboard.test.ts` covering the stable-value/label decoupling and the offline-record `stale` case.

The colour mapping is unchanged; borders render byte-identical and the visible status text stays fully translated. Only the lookup key changed from translatable text to a language-independent value.

### On the CSS-module migration (tradeoff)
SCSS can't import `theme.ts`, and the repo's `theme.scss` has drifted from it (`$warning` `#efda70` vs the JS `#FEDD2F`, and there is no token for the `ok` green `#298217`). To avoid duplicating or hardcoding those colours into SCSS, the colour map stays in JS (sourced from `theme.color`) and is passed to the module via a CSS custom property. This is a deliberate hybrid: static styling lives in the SCSS module, the dynamic token stays in JS, and `theme.ts` remains the single source of truth.

### Deferred: TypeScript migration of the larger touched files
`Collect.jsx` and `CollectRecordsMixin.js` are intentionally left as JS for this fix:
- Neither uses PropTypes, so the checklist's PropTypes-to-interface driver does not apply.
- `CollectRecordsMixin.js` is one of 11 sibling mixins that are all `.js`; migrating one in isolation is inconsistent, and the `(Base) => class extends Base` pattern needs base-state typing to compile cleanly.
- Renaming `Collect.jsx` to `.tsx` type-checks with roughly zero new annotations, but surfaces two latent errors from untyped `styled(...)` exports in the shared `table.js`. Resolving them properly means typing `table.js` (a ~350-line shared module), which is out of scope here.

These belong in a dedicated TypeScript pass, starting with `table.js`.

### Testing
New and existing Collect tests pass; ESLint and `tsc --noEmit` are clean. The new test was verified to fail against the unfixed code, confirming it guards the regression. Border rendering and language switching were confirmed at runtime.

Before:
image

After:
image

---Every PR---

---Transitional Changes---

- [x] Touched JS files are migrated to TypeScript where in scope; PropTypes are replaced with TypeScript interfaces in converted files and not added for shapes already typed in `mermaidDataTypes.ts`. (`collectConstants.ts` done; `Collect.jsx` and `CollectRecordsMixin.js` deferred, see above.)

---Post-Merge---

- [ ] Code has merged and build success is confirmed in &#39;Actions&#39;

## Summary by CodeRabbit

- **New Features**
  - Collect records now use consistent validation status indicators, including a clear fallback for missing or unrecognized statuses.
  - Status colors and visual markers in the Collect table are aligned with each record’s normalized validation state.

- **Bug Fixes**
  - Records created offline or with unexpected validation values are now displayed with the appropriate stale status instead of inconsistent labels or styling.

- **Tests**
  - Added coverage for normalized validation statuses and offline records.
  - Corrected test setup issues to improve reliability.


src/App/mermaidData/databaseSwitchboard/tests/collectRecordStatusValue.databaseSwitchboard.test.ts
```diff
@@ -0,0 +1,62 @@
+import { expect, test } from 'vitest'
+import { getDatabaseSwitchboardInstanceAuthenticatedOnlineDexieSuccess } from './testHelpers.DatabseSwitchboard'
```
tienhoah: `DatabseSwitchboard` looks like a typo for `DatabaseSwitchboard`. 
If this is also the pre-ixisting filename, can you @JBurkinshaw  change it?

tienhoah (changes requested):

src/App/mermaidData/databaseSwitchboard/tests/collectRecordStatusValue.databaseSwitchboard.test.ts
JBurkinshaw: Yes, this is pre-existing. I just pushed an update to fix it

src/App/mermaidData/databaseSwitchboard/tests/collectRecordStatusValue.databaseSwitchboard.test.ts
```diff
@@ -0,0 +1,62 @@
...
+    // ...and never the translated display text the colours used to be keyed on.
+    expect(record.uiLabels.statusValue).not.toEqual(record.uiLabels.status)
+  })
+
+  // guard the scenario the bug was about: at least one record with a real status
+  // whose translated label differs from its stable value.
+  expect(records.some((record) => record.uiLabels.statusValue === VALIDATION_STATUS.error)).toBe(
+    true,
+  )
+})
```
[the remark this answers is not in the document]
JBurkinshaw: Not making this change: the test i18n instance is mocked in `cimode` with no translation resources loaded (`setupTests.js`), so `t()` returns the translation key rather than localized text regardless of locale. Switching language wouldn't produce a translated label to assert against, and coupling to translated strings runs against the harness's deliberate "assert on stable keys, not English text that may change" pattern.

The existing assertion already covers the regression: `statusValue` (`'error'`) is proven not equal to the label, confirming the colour key is decoupled from the (translatable) status text. Test passes as-is.

src/components/pages/Collect/Collect.jsx
```diff
@@ -45,9 +45,20 @@ import {
...
+import theme from '../../../theme'
+import { VALIDATION_STATUS } from './collectConstants'
+import styles from './Collect.module.scss'
 import { getIsEmptyStringOrWhitespace } from '../../../library/getIsEmptyStringOrWhitespace'
 
+// Keyed on the stable status value (not the translated label) so the status
+// border colour is consistent across all app languages.
+const RECORD_STATUS_COLORS = {
+  [VALIDATION_STATUS.error]: theme.color.cautionColor,
+  [VALIDATION_STATUS.ok]: '#298217',
```
tienhoah: can we add this color to theme.ts

tienhoah (changes requested):
1 small change

src/components/pages/Collect/Collect.jsx
JBurkinshaw: Good idea. This colour is pre-existing (it came over unchanged from the old `Collect.styles.js` map. I have added a `successColor` token to `theme.ts` and pointed the map at it. Value is unchanged, so the border renders identically.

tienhoah (approved):

PR #1666, opened 2026-07-16
feat: [M1981] Add columns to GFCR facility / solution table

JBurkinshaw:
See M1981 trello card for more discussion around this.

Adds columns to the facility / solution table (Collect / GFCR) and refines how per-type fields are displayed.

The originally requested fields (Geographical coverage, Name of TAF (incubator), and Total number of solutions supported) already existed on the add/edit form and are persisted with the indicator set; this change surfaces them in the table. Columns are sortable and included in the search filter. No model, form, or API changes were needed.

### Column display

Following discussion on the ticket, fields that do not apply to a row's solution type now render **blank** rather than a cross, "None" or `0`:

- Gender 2X criteria and Local enterprise show only for the types they apply to.
- Used a TAF (incubator)? keeps "None" where an applicable type answered no, but is blank for types the question doesn't apply to.
- Total number of solutions supported shows its value (including a genuine `0`) only for applicable types.

Each of these columns, alongside the existing text columns, is **hidden when no row in the set can populate it**.

Field applicability now lives in a shared `financeSolutionFieldVisibility` module used by both the table and the edit modal, so the two agree on which fields apply to which type.

### Notes

- Existing `Type` column is kept; Local enterprise is ordered before Gender 2X to match the modal.
- Headers use the wording confirmed on the ticket ("Total number of solutions supported", "Name of TAF (incubator)").
- Renames the `number_of_solutions_supported_by_table_header` token to drop the redundant `_by` fragment.

### Testing

Manually verified by adding solutions of each type: non-applicable cells render blank, a real `0` still shows for applicable types, and each applicability-driven column hides when no row can populate it. Lint passes.

---Every PR---

- [ ] Changes have been tested locally
- [ ] Lint has run and any errors have been resolved
- [ ] Prettier has run on any changed files
- [ ] Unit tests have run and passed

---Transitional Changes---

---Post-Merge---

### DIFF
--- a/src/locales/en/translation.json
+++ b/src/locales/en/translation.json
@@ -352,6 +352,7 @@
         "name_helper": "Enter the name of each business or financial mechanism solution financed by the GFCR Programme separately. You may copy solutions from previous reports. Indicators F8, F9, and F10 will be disaggregated by both solution and sector. New solutions financed by the Programme will be added here progressively.<br /><gfcrPdfLink>View M&E Toolkit PDF</gfcrPdfLink>",
         "no_finance_solutions": "No facilities / solutions yet",
         "number_of_solutions_supported_by": "Number of solutions supported by",
+        "number_of_solutions_supported_by_table_header": "Total number of solutions supported",
         "pcf_revenues_error": "Remove revenues before changing type to Programmatic co-financing",
         "save": "Save facility / solution",
         "sector": "Sector",
--- a/src/components/pages/gfcrPages/GfcrIndicatorSetForm/subPages/FinanceSolutions.jsx
+++ b/src/components/pages/gfcrPages/GfcrIndicatorSetForm/subPages/FinanceSolutions.jsx
@@ -44,7 +44,7 @@ const FinanceSolutions = ({ indicatorSet, setIndicatorSet, choices, displayHelp 
   const geographicalCoverageHeaderText = t('gfcr.forms.finance_solutions.geographical_coverage')
   const tafNameHeaderText = t('gfcr.forms.finance_solutions.taf_name')
   const numberOfSolutionsSupportedByHeaderText = t(
-    'gfcr.forms.finance_solutions.number_of_solutions_supported_by',
+    'gfcr.forms.finance_solutions.number_of_solutions_supported_by_table_header',
   )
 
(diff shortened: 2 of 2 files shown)


src/locales/en/translation.json
```diff
@@ -352,6 +352,7 @@
         "name_helper": "Enter the name of each business or financial mechanism solution financed by the GFCR Programme separately. You may copy solutions from previous reports. Indicators F8, F9, and F10 will be disaggregated by both solution and sector. New solutions financed by the Programme will be added here progressively.<br /><gfcrPdfLink>View M&E Toolkit PDF</gfcrPdfLink>",
         "no_finance_solutions": "No facilities / solutions yet",
         "number_of_solutions_supported_by": "Number of solutions supported by",
+        "number_of_solutions_supported_by_table_header": "Total number of solutions supported",
```
tienhoah, line 355: Nitpick: The token name should closely resemble the value. Since the value is `Total number of solutions supported`, consider renaming the key to something cleaner like `total_solutions_supported` or `total_solutions_supported_header` to remove the trailing `_by`.

tienhoah (changes requested):

src/locales/en/translation.json
JBurkinshaw: Good catch with the `_by` reading strangely. I'm going push back on the rename itself though

Our `docs/TranslationBestPractices.md` actually steers away from matching the key to the value:
- 2: "Token names should convey their purpose without relying on the actual text content."
- 4: "Avoid Embedding Actual Text in Keys... this practice hinders flexibility and can complicate future text changes."

`total_solutions_supported` is basically a transcription of the display string, so if the copy changes later the key becomes misleading. I'd rather keep `number_of_solutions_supported...` so it stays grouped with its sibling token and the `_table_header` suffix signals its purpose.

That said, I agree that `_by` is not needed It only exists because the sibling `number_of_solutions_supported_by` is a sentence fragment ("Number of solutions supported by") expecting something appended, whereas this value is a standalone sentence. I'll drop it:
`number_of_solutions_supported_by_table_header` → `number_of_solutions_supported_table_header`

Keeps the grouping and purpose-based naming, and removes the `_by` you flagged.

tienhoah (approved):

PR #1674, opened 2026-07-22
feat: [M1947] Copy facility / solution

JBurkinshaw:
## Summary

Adds a "Copy facility / solution" feature to the GFCR Facilities/solutions tab, letting users copy existing facilities/solutions (with their investments and revenues) from other fund indicator sets in the same project into the current one. Also migrates the touched view to TypeScript and CSS modules per the transitional-change conventions.

> **Retargeted to `develop`.** This PR previously targeted `feat/gfcr/feature-branch`. That branch turned out to hold no unique work (no non-merge commits vs `develop`, and byte-identical in the GFCR area), so it was not a useful integration point. The branch has been rebuilt directly on `develop` and the base changed accordingly. Details in "Rebuild notes" below.

### Feature: copy facility / solution
- New **Copy facility / solution** button on the Facilities/solutions tab, next to Add.
- Opens a modal (`CopyFinanceSolutionsModal`) listing every facility/solution from the project's *other* fund indicator sets, with search, sort, pagination, row selection, and a "view selected only" filter.
- Selected rows are deep-cloned into the current indicator set along with their nested investments and revenues. All `id` fields, and the `finance_solution` back-references on investments and revenues, are stripped via `stripId` so the records save as new (`copyHelpers.ts`).
- The trigger button is disabled with an explanatory tooltip when no other indicator set in the project has anything to copy, matching the existing disabled-button pattern used for adding revenues.
- New i18n tokens for the modal, button, tooltip, and toasts.

### Refactor: TypeScript + CSS modules (transitional changes)
- Migrated `FinanceSolutions.jsx` to `.tsx`: PropTypes replaced with a TS interface, reusing `Choices` from `mermaidDataTypes.ts` and the shared finance-solution types (`FinanceSolution`, `InvestmentSource`, `Revenue`, `IndicatorSet`) promoted into `mermaidDataTypes.ts` (also consumed by the modal).
- Moved the toolbar button layout into a co-located CSS module (`FinanceSolutions.module.scss`) so the Add and Copy buttons sit side by side, replacing `StyledToolbarButtonWrapper` and avoiding changes to the shared styled-component.
- The new modal styling uses a CSS module (`CopyFinanceSolutionsModal.module.scss`); minimal type declarations were added to `table.d.ts` so the TSX modal type-checks against the existing shared table styles.
- Tokenised the hardcoded `'None'` incubator label as `gfcr.forms.finance_solutions.no_incubator`.

### Incidental bug fix
- `GfcrGenericTable` expects `searchFilteredRowsLength`, but `Revenues`, `Investments` and `FinanceSolutions` all passed `searchFilteredRowLength`, so the "X of Y" filtered-row count did not display during search. Fixed in all three (separate commit).
- Connecting that prop exposed a second bug underneath: the count mapped filtered rows back to source records by a key the table cell data lacks (`row.original.id`, `row.original.name`), so it always resolved to zero during a search, which also collapsed the page-size dropdown to a single `0` option. All four GFCR tables now take the count from react-table's `rows`, which already holds the filtered set, removing the local state and the lookup. `Gfcr` is included because it had both halves of the bug, so its indicator-sets footer now shows a filtered count for the first time.

## Rebuild notes

The branch was originally cut from `develop` before M1981 landed, so the TypeScript conversion was made against the pre-M1981 `FinanceSolutions.jsx`. Merging that into current `develop` produced a rename-detected conflict whose obvious resolution would have silently reverted M1981: specifically, restoring `IconCheckLabel` elements as the `gender_smart` and `local_enterprise` cell values, which is exactly what M1981's `fix: Sort yes/no finance columns by value` removed because `reactTableNaturalSort` then compares every row as `[object Object]`.

Rather than resolve that conflict, the conversion was redone on top of current `develop`. M1981's behaviour is preserved unchanged:
- `columnHasData` column filtering and the per-type applicability gates are untouched.
- `gender_smart` and `local_enterprise` keep raw `boolean | null` cell values feeding their `Cell` renderers, so sorting still works.
- The `geographical_coverage`, `taf_name` and `number_of_solutions_supported_by` columns are retained.

The rename-detected diff on `FinanceSolutions` reports 83% similarity and none of the above appears in it, which is the check that they are unmodified.

The pre-rebuild tip is preserved on the remote as `feat/gfcr/M1947-copy-facility-solution-pre-rebuild` (`f1a4d7bc`).

## Phase 4 containment

This PR contains **no Phase 4 enforcement changes**. Verified:
- `FinanceSolutionModal.jsx`, `ReportTitleAndDateForm.jsx` and `GfcrIndicatorSet.jsx` are untouched.
- The Phase 2 non-standard-value fallbacks are intact and identical to `develop` (`isStandardFsType` guard, `gfcr.non_standard_title_suffix`, the `NonStandard` title branch).
- `translation.json` has additions only, no deletions.
- M1971 / PR #1677 is not an ancestor of this branch. #1677 now targets `develop` independently and must not merge until the API Phase 4 release.

On the data-migration question: the copy is a verbatim clone that only strips ids, so it neither adds nor requires any field and cannot bring Phase 4 enforcement forward. It is, however, the only write path that bypasses `FinanceSolutionModal`, which already rejects an empty or non-standard `fs_type` and blanks non-applicable fields on submit. So it can clone a record that has not yet been migrated. That is covered by running the Phase 3 audit immediately before the Phase 4 release rather than earlier.

## Testing

`tsc`, `eslint` and `prettier` are clean. Full unit suite passes (646 passed, 7 skipped). Production build succeeds.

The feature code (`CopyFinanceSolutionsModal`, `copyHelpers`) is unchanged from the version that was manually QA'd before the rebuild:
- Trigger button: disabled with tooltip when the project has no other indicator sets, or when none of them have facilities/solutions; enabled and side-by-side with Add otherwise.
- Modal: filtering by indicator set/solution name, "view selected only", pagination, sorting, and single/multi-row selection all work against facilities/solutions sourced from other indicator sets only.
- Copy action: selected rows (single and multiple, including across different source indicator sets) save as new records with fresh ids; copied investments/revenues correctly reference the new facility/solution, not the source. Source indicator set is left untouched.
- Cancel discards the selection without copying anything.
- Simulated a failed save (blocked request): error toast shown, modal stays open with selection intact.

⚠️ **`FinanceSolutions` itself was rebuilt after that QA round and has not been re-verified in the browser.** Worth a pass over the Facilities/solutions tab before merge, in particular that the Gender 2X criteria and Local enterprise columns still render icons and still sort, and that column blanking still behaves per M1981.

## Translations

English is the source of truth (`src/locales/en/translation.json`); syncs to Lokalise on merge to `develop`.

**Added:**

| Key | Value |
|---|---|
| `filters.by_indicator_set_or_solution_name` | "Filter by indicator set or solution name" |
| `gfcr.forms.finance_solutions.copy` | "Copy facility / solution" |
| `gfcr.forms.finance_solutions.copy_info` | "Related Investments and Revenues will be copied as well." |
| `gfcr.forms.finance_solutions.copy_selected` | "Copy selected facilities / solutions" |
| `gfcr.forms.finance_solutions.indicator_set` | "Indicator set" |
| `gfcr.forms.finance_solutions.no_copy_targets` | "There are no other fund indicator sets in this project with facilities / solutions to copy" |
| `gfcr.forms.finance_solutions.no_incubator` | "None" |

**Removed:** none

## Notes for reviewer

- **Copy modal column set.** The modal uses the column list from the M1947 ticket and does not apply M1981's per-type blanking, so it shows "None" and check/cross icons on every row where the main table now blanks those columns. This matches the ticket's stated columns but the two tables will look inconsistent. Flagging for a product call: leave and raise a follow-up, or fold the applicability gating in here.
- **Modal state on dismiss.** `onDismiss` only closes the modal, so react-table's selection and the "view selected only" toggle survive and are still there on reopen. Cancel does not copy anything, so this is cosmetic, but `toggleAllRowsSelected(false)` on dismiss would be tidier.
- **View-selected-only pagination.** `pageCount` is overridden for the selected-rows view but `canNextPage` / `canPreviousPage` still come from the underlying pagination state, so paging can be inconsistent in that mode. Inherited from `CopySitesModal`.
- **Gfcr footer is newly visible.** Fixing the prop name on `Gfcr` means the indicator-sets table shows "showing N of X filtered from Y" during search where it previously showed nothing. Intended, but it is new UI on a page M1947 does not otherwise touch, so worth a look in QA.
- **Error path** fires both `toast.error` and `handleHttpResponseError`, which is likely the "server unreachable" cascade noted during QA. Also present on the existing Add flow.

**Pre-existing issues surfaced during QA, not introduced by this PR:**
- Column sorting only adds a clicked column as a secondary tie-breaker rather than replacing the active sort, because `isMultiSortEvent` is hardcoded to always return true (also present in `CopySitesModal`'s Exposure column). Raising separately for a product/eng call on desired behaviour.

## Out of scope
- The shared design-system styled-components this view consumes (buttons, table primitives, modal, layout helpers) are used app-wide and are not migrated here. Full GFCR CSS-module conversion is captured as a separate tech-debt card.

---

**Every PR**
- [x] Unit tests have run and passed *(no tests cover this GFCR area; full suite green)*

**Transitional Changes**
- [ ] Any language.js tokens no longer used are removed *(n/a, only additions)*
- [ ] Any necessary updates to the documentation have been made *(n/a)*
- [ ] Unit tests have been added or updated *(n/a, no existing coverage for this area)*
- [x] styled-components code in changed files has been updated to CSS modules in the styles folder *(new/changed styling only; shared components out of scope)*
- [x] Touched JS files are migrated to TypeScript where in scope; PropTypes replaced with TS interfaces, reusing `mermaidDataTypes.ts`

**Post-Merge**
- [ ] Merge `develop` into `feat/gfcr/feature-branch` to align it
- [ ] Delete `feat/gfcr/M1947-copy-facility-solution-pre-rebuild` once happy

## Summary by CodeRabbit

* **New Features**
  * Added the ability to copy finance solutions from other indicator sets.
  * Added search, sorting, selection, “view selected only,” and pagination options to the copy dialog.
  * Added translated labels and messages for finance solution copying and empty states.

* **Bug Fixes**
  * Corrected filtered-row counts in investment, revenue, and finance solution tables.
  * Improved empty-value labeling and modal state handling.

* **Style**
  * Added consistent spacing and layout styling for finance solution controls and dialogs.


src/components/pages/gfcrPages/GfcrIndicatorSetForm/modals/CopyFinanceSolutionsModal.tsx
```diff
@@ -0,0 +1,454 @@
...
+
+      setIndicatorSet(response)
+      toast.success(...getToastArguments(indicatorSetSaveSuccessText))
+      toggleAllRowsSelected(false)
+      setIsSaving(false)
+      onDismiss()
+    } catch (error) {
+      setIsSaving(false)
+
+      if (error) {
```
tienhoah: **Nit**: `if (error)` inside a `catch` block is always true — a caught value is by definition truthy, so the condition never gates anything. `toast.error` and `handleHttpResponseError` can be called directly without the wrapper.

src/components/pages/gfcrPages/GfcrIndicatorSetForm/modals/CopyFinanceSolutionsModal.tsx
```diff
@@ -0,0 +1,454 @@
...
+
+  const _setFilterPrefs = useEffect(() => {
+    handleSetTableUserPrefs({ propertyKey: 'globalFilter', currentValue: globalFilter })
+  }, [globalFilter, handleSetTableUserPrefs])
+
+  const _resetToPageOneWhenViewSelectedRowsIsOn = useEffect(() => {
+    if (isViewSelectedOnly) {
+      gotoPage(0)
+    }
+  }, [isViewSelectedOnly, gotoPage])
```
tienhoah, line 290: **Nit**: is this still the preferred pattern to call prefix variable for useEffect that we want to keep going forward, or are we moving away from it.

tienhoah (changes requested):
Looks good, 1 small change and 1 question

src/components/pages/gfcrPages/GfcrIndicatorSetForm/modals/CopyFinanceSolutionsModal.tsx
JBurkinshaw: You're right. I removed it so `toast.error` / `handleHttpResponseError` is called directly directly

src/components/pages/gfcrPages/GfcrIndicatorSetForm/modals/CopyFinanceSolutionsModal.tsx
JBurkinshaw: It's the existing convention. It's used all over the codebase and this modal matches `CopyManagementRegimesModal`, which uses the same naming. I kept it for consistency rather than introducing a different style here. Happy to revisit whether we move away from it, but I'd suggest doing that as its own discussion rather than in this PR. Personally I like this convention as it makes it easy to determine what the useEffect's intent is from the const name.

tienhoah (approved):

src/components/pages/gfcrPages/GfcrIndicatorSetForm/modals/CopyFinanceSolutionsModal.tsx
```diff
@@ -0,0 +1,455 @@
...
+      <CopyModalPaginationWrapper>
+        <PageSelector
+          onPreviousClick={previousPage}
+          previousDisabled={!canPreviousPage}
+          onNextClick={nextPage}
+          nextDisabled={!canNextPage}
+          onGoToPage={gotoPage}
+          currentPageIndex={pageIndex}
+          pageCount={pageCount}
+        />
```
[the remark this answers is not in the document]
JBurkinshaw: Confirmed, and reachable two ways: with one row selected in a multi-page table "Next" stays enabled and pages to an empty selected-only view, and toggling the filter on while already past page 1 lands on an empty view with no click at all.

Leaving it here. This is already flagged under "Notes for reviewer" in the PR description: the selected-only view is lifted from `CopySitesModal`, which has the same wiring, so I'd rather fix both in one follow-up than diverge the two modals in this PR. Cancel copies nothing, so the impact is a confusing empty table rather than bad data.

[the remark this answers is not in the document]
JBurkinshaw: Fixed

src/components/pages/gfcrPages/GfcrIndicatorSetForm/subPages/FinanceSolutions.tsx
```diff
@@ -375,7 +425,7 @@ const FinanceSolutions = ({ indicatorSet, setIndicatorSet, choices, displayHelp
       onPageSizeChange={handleRowsNumberChange}
       pageSize={pageSize}
       unfilteredRowLength={indicatorSet.finance_solutions.length}
-      searchFilteredRowLength={searchFilteredRowsLength}
+      searchFilteredRowsLength={searchFilteredRowsLength}
```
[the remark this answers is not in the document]
JBurkinshaw: Correct, and reachable two ways: with one row selected in a multi-page table "Next" stays enabled and pages to an empty selected-only view, and toggling the filter on while already past page 1 lands on an empty view with no click at all.

Leaving it here. This is already flagged under "Notes for reviewer" in the PR description: the selected-only view is lifted from `CopySitesModal`, which has the same wiring, so I'd rather fix both in one follow-up than diverge the two modals in this PR. Cancel copies nothing, so the impact is a confusing empty table rather than bad data.

PR #1689, opened 2026-07-27
2026 secure actions

ms280690:
https://trello.com/c/w96jL1Xi

## Summary by CodeRabbit

- **Chores**
  - Improved release, PR, and deployment workflow reliability with clearer version handling and tighter permissions.
  - Strengthened automation security by hardening runners, pinning action versions, and explicitly controlling secret availability.
  - Added/updated security and YAML linting workflows for GitHub Actions.
- **Documentation**
  - Clarified translation review guidance and refined release/translation sync workflow instructions.
- **Tests**
  - Improved integration test stability for observation/popup flows and adjusted test mocks for consistent assertions.


michaelconnor00 (commented):
All the github workflow changes look good. 

I cannot speak for the tsx/jsx changes. Those seem to be in this PR by mistake...

JBurkinshaw (changes requested):
25 files changed here. It looks like a lot of them were just formatted. Can you remove the files where only formatting changed from the PR so it is easier to review?

ms280690:
> 25 files changed here. It looks like a lot of them were just formatted. Can you remove the files where only formatting changed from the PR so it is easier to review?

Done

.github/workflows/deployment.yml
```diff
@@ -29,16 +22,28 @@ jobs:
...
     uses: ./.github/workflows/_cdk.yml
-    secrets: inherit
+    secrets:
+      VITE_AUTH0_CLIENT_ID: ${{ secrets.VITE_AUTH0_CLIENT_ID }}
+      VITE_MAPBOX_ACCESS_TOKEN: ${{ secrets.VITE_MAPBOX_ACCESS_TOKEN }}
+      VITE_CORAL_ATLAS_APP_ID: ${{ secrets.VITE_CORAL_ATLAS_APP_ID }}
+      VITE_SENTRY_DSN: ${{ secrets.VITE_SENTRY_DSN }}
+      SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
+      CDK_DEFAULT_ACCOUNT: ${{ secrets.CDK_DEFAULT_ACCOUNT }}
+      CDK_DEPLOY_ROLE: ${{ secrets.CDK_DEPLOY_ROLE }}
```
JBurkinshaw: Swapping `secrets: inherit` for an explicit list changes where the secrets get looked up. They're now resolved here in `deployment.yml`, and this job doesn't set an `environment:`, so it only sees repo-level secrets.

In the repo settings:
- Repo level: `CDK_DEFAULT_ACCOUNT`, `CDK_DEPLOY_ROLE`, `SENTRY_AUTH_TOKEN`, `VITE_SENTRY_DSN`, `VITE_AUTH0_CLIENT_ID`
- Only on the `dev` and `Production` environments: `VITE_MAPBOX_ACCESS_TOKEN`, `VITE_CORAL_ATLAS_APP_ID`

So lines 41 and 42 are looking up names that don't exist at this level, and I'd expect them to come through as empty strings. `required: true` won't catch it, as it only checks the line is there, not that the value isn't blank. Line 40 is a slightly different problem: `VITE_AUTH0_CLIENT_ID` exists in both places, so it'd get the repo-level value instead of the environment's.

If that's what's happening, a dev or prod deploy ships a build with no Mapbox token and no Coral Atlas app ID. Same thing on lines 26-27 for the test job.

The reason I'm not certain is that the `_cdk.yml`'s job does set `environment:`, so environment secrets are in scope inside it too, and I don't know for sure which value wins when a secret that is passed in and an environment secret share a name. If environment wins, this will be fine and the explicit list is just redundant.

Worth confirming before this merges, because PR CI won't catch it. `_test.yml` isn't called from `pr.yml` and this job only runs on push to develop, so the first time we'd find out is after merge. The quickest check would be a temporary step in `_cdk.yml` doing `[ -z "$VITE_MAPBOX_ACCESS_TOKEN" ] && echo EMPTY` and pushing to develop once. If it is broken, adding `environment:` to these caller jobs should fix it.

.github/workflows/pr.yml
```diff
@@ -144,12 +152,15 @@ jobs:
...
         if: always() && steps.cdk-diff.outcome != 'skipped'
-        uses: actions/github-script@v7
+        env:
+          WORKFLOW_NAME: ${{ github.workflow }}
+          CDK_DIFF_OUTCOME: ${{ steps.cdk-diff.outcome }}
+        uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9.0.0
         with:
           script: |
             const fs = require('fs');
             const log = fs.readFileSync('/tmp/cdk-diff.log', 'utf8');
```
JBurkinshaw: This reads the raw diff log off disk and posts it to the PR. Secret hiding only applies to what gets printed in workflow logs, not to what we send through the API, so anything in that log goes into the comment as-is.

The result is on this PR right now. The `cdk diff` comment contains:
```
start: Publishing dev-webapp Template (554812291621-us-east-1-a0e639cd)
```

That's the AWS account ID, and this repo is public. Not a credential so not worth panicing about, but we store it as `CDK_DEFAULT_ACCOUNT` in secrets, so we're treating it as sensitive in one place and publishing it in another. It also stays in the comment history after the PR closes.

Could you potentially replace the account ID with a placeholder before this step reads it?

JBurkinshaw (approved):

michaelconnor00 (approved):

PR #1690, opened 2026-07-27
fix: [M2044] GFCR numeric input consistency — replace custom handlers with Mantine NumberInput

tienhoah:
## What changed:

- Installed Mantine v9 (`@mantine/core`, `@mantine/hooks`) alongside existing MUI — no peer dep conflicts
- Added `getBrowserLocale` helper (`src/library/numbers/`) that reads `navigator.language` and resolves it via `Intl.NumberFormat` with an `en-US` fallback
- Built `GfcrNumberInput` component wrapping Mantine's `NumberInput`: locale-aware decimal/thousands separators, rounds to N decimal places on blur, clamps min/max, blocks negatives, no steppers, scroll wheel disabled
- Migrated `GfcrDecimalInputField` from `.jsx` to `.tsx` using `GfcrNumberInput` — fixes comma-decimal locales storing `15` instead of `1.5`, and empty fields now saving `null` instead of restoring initial value
- Replaced native `type="number"` inputs in `InvestmentModal` and `RevenueModal` with `GfcrNumberInput` (USD $ unit left-aligned, 2 decimal places) — removes browser steppers and inconsistent scroll behaviour
- Retired `formikHandleDecimalInputOnBlur` (no longer used; `formikHandleGfcrNumberInputChange` kept — still used by `GfcrIntegerInputField`)
- Added `renderInput` prop to `InputNoRowWithLabelAndValidation` to allow full input replacement while preserving label/validation chrome — non-breaking, existing callers unaffected
- Wrapped test providers in `MantineProvider` so `NumberInput` renders correctly in Vitest
- Added no-nested-ternary convention to `.claude/CLAUDE.md`

---Every PR---

---Transitional Changes---

---Post-Merge---

## Summary by CodeRabbit

* **New Features**
  * Added locale-aware number inputs with decimal precision, minimum/maximum limits, negative-value controls, units, and improved empty-value handling.
  * Updated investment, revenue, and GFCR decimal fields to use the new number input experience.
  * Added support for custom input rendering and improved accessibility attributes.

* **Bug Fixes**
  * Improved numeric validation, rounding, clamping, and integer-only input behavior.
  * Added reliable locale handling for decimal separators.

* **Tests**
  * Added coverage for number input behavior, locale formatting, validation, and browser compatibility.


src/components/pages/gfcrPages/GfcrIndicatorSetForm/modals/InvestmentModal.jsx
```diff
@@ -293,21 +293,25 @@ const InvestmentModal = ({
...
-                event,
-                fieldName: 'investment_amount',
-              })
+            renderInput={
+              <GfcrNumberInput
+                id="investment-amount-input"
+                name="investment_amount"
+                aria-labelledby="aria-labelinvestment-amount-input"
+                aria-describedby="aria-descpinvestment-amount-input"
+                value={formik.values.investment_amount}
```
JBurkinshaw: 🤖 *AI code review (Claude)* · **major**

**Focus and blur on an untouched amount rewrites the stored value and marks the form unsaved.**

The API returns `investment_amount` as a string (`mermaidDataTypes.ts:424`), and this passes it straight into `GfcrNumberInput`, which expects `number | null`. `GfcrDecimalInputField.tsx:27-31` guards with a `parseFloat`; the modals don't.

Mantine's blur handler sanitises a string into a number, and since `trimLeadingZeroesOnBlur` defaults to `true`, `_value !== sanitizedValue` is always true for a string, so `setValue` fires `onChange` (`NumberInput.mjs:316-330`). That sets `formik.dirty`, which line 218 feeds into the Save button state.

Repro:
1. Save an investment with amount `1234.56`, then reload.
2. Reopen it, click into the amount and tab out without typing.
3. Save switches to unsaved changes, and saving persists a re-rounded value.

Verified against this head: mounting with `"1234.567"` then focus/blur emits `onChange(1234.567)` followed by `onChange(1234.57)`, with no user input.

Fix: parse at the boundary in both modals. With three call sites, worth extracting the parse out of `GfcrDecimalInputField` into a shared helper.

```diff
-                value={formik.values.investment_amount}
+                value={parseGfcrNumber(formik.values.investment_amount)}
```

`RevenueModal.jsx:309` needs the same change.

Also worth a `typeof value === 'number'` guard on `roundToDecimalPlaces` (`GfcrNumberInput.tsx:98`). It is safe today only because Mantine normalises before calling `onBlur?.(event)`.

src/library/numbers/getBrowserLocale.ts
```diff
@@ -0,0 +1,8 @@
+export const getBrowserLocale = (): string => {
+  const locale = navigator.language ?? 'en-US'
+  try {
+    return new Intl.NumberFormat(locale).resolvedOptions().locale
```
JBurkinshaw: 🤖 *AI code review (Claude)* · **minor**

**Doesn't resolve to a region, which requirement 6 asks for.**

`Intl.NumberFormat` doesn't expand a bare language tag. `new Intl.NumberFormat('de').resolvedOptions().locale` returns `'de'`, not `'de-DE'`. Same for `en`, `id` and `fr`.

Harmless in practice, since Intl picks sensible separators for bare tags and browsers usually report a full tag. But the card specifies a region-resolved locale, the unit test passes only because jsdom reports `en-US`, and this helper is the seam the follow-up locale ticket plugs into.

`Intl.Locale.maximize()` adds the region. Keep the `NumberFormat` call for validation, which `Intl.Locale` alone doesn't give you: `new Intl.Locale('not-a-locale')` yields `not-PE` rather than throwing.

```ts
const resolved = new Intl.NumberFormat(locale).resolvedOptions().locale
const { language, region } = new Intl.Locale(resolved).maximize()
return region ? `${language}-${region}` : resolved
```

Checked against `en-US`, `de`, `id`, `not-a-locale`, `en_US`, `undefined` and `''`: gives `en-US`, `de-DE`, `id-ID`, then `en-US` for the rest. Both existing tests stay green.

JBurkinshaw (commented):
🤖 Automated code review by Claude. 3 inline comments, ordered by severity. Checked against this head, Mantine's source in `node_modules`, and the GFCR models in `mermaid-api`. Suggestions to weigh, not blockers.

Earlier findings on `null` handling, the missing aria associations, and the `f4_1`/`f4_3` decimal places are fixed by 8f2891b8 and are not repeated here.

Two things outside the diff. The AC asks for a side-by-side screenshot, which isn't in the description yet. And nothing exercises `GfcrDecimalInputField` or the two modals, which is why the first finding below went unnoticed. The full suite is green on this head: 702 passed, 7 skipped.

src/components/pages/gfcrPages/GfcrIndicatorSetForm/GfcrDecimalInputField.tsx
```diff
@@ -0,0 +1,68 @@
...
+        <GfcrNumberInput
+          id={id}
+          aria-labelledby={`aria-label${id}`}
+          aria-describedby={`aria-descp${id}`}
+          value={numericValue}
+          onChange={(val) => formik.setFieldValue(id, val)}
+          onBlur={handleBlur}
+          onFocus={handleInputFocus}
+          decimalPlaces={maxNumberOfDecimals}
+          min={0}
```
JBurkinshaw: 🤖 *AI code review (Claude)* · **minor**

**No GFCR field sets a `max`, so an over-long entry reaches the API as a raw 400.**

`GfcrNumberInput` supports `max` and clamps on blur, but no call site passes one. Every field is capped server-side by `max_digits` in `mermaid-api` `models/gfcr.py`:

| Field | API cap |
|---|---|
| `f4_1`, `f4_2` (%) | 999.9 |
| `f4_3` (kg/ha) | 9999.9 |
| `f1_1`, `f2_*`, `f3_1` (km²) | 999999.99999 |
| `investment_amount` | 9999999999.99 |
| `revenue_amount` | 999999999.99 |

Repro: type `1000` into F 4.1 and save. The API returns "Ensure that there are no more than 4 digits in total", with no field-level feedback in the UI.

The percentage fields are the sharpest case, where the sensible cap is 100 rather than the API's 999.9.

This leaves the card's "min/max enforced consistently" half done. Fix: add a `maxValue` prop alongside `maxNumberOfDecimals` and set it per field. Pre-existing rather than a regression, so a follow-up ticket is reasonable if you'd rather keep this PR tight.

src/components/pages/gfcrPages/GfcrIndicatorSetForm/modals/InvestmentModal.jsx
JBurkinshaw: Fixed in 617c5bd, though at the source rather than the `value=` prop. Both `*InitialValues.js` files now run the amount through a new `parseGfcrNumber` helper, so formik holds a number from the moment the modal opens and Mantine has nothing left to coerce. Patching `value=` alone would have stopped the phantom change but left a string sitting in formik, which then goes out in the payload.

Also added the `typeof` narrowing on the rounding path, plus a regression test that focuses and blurs an untouched value and asserts `onChange` never fired.

src/library/numbers/getBrowserLocale.ts
JBurkinshaw: Fixed in 617c5bd, using `Intl.Locale.maximize()` with the `NumberFormat` call kept for validation. Added tests for `de` resolving to `de-DE` and for `en-GB` keeping the region it already has.

src/components/pages/gfcrPages/GfcrIndicatorSetForm/GfcrDecimalInputField.tsx
JBurkinshaw: Fixed in 617c5bd. New `maxValue` prop, plus a `gfcrFieldMaximums.ts` documenting each ceiling against `max_digits` on the models. All 13 decimal fields and both modal amounts now pass one, with cover percentages capped at 100.

Manually testing found one issue. Clamping to a ceiling and then rounding can push it back over. F 1.1 allows 2 decimals but the km² ceiling is `999999.99999`, which rounded to `1000000` and 400s with "no more than 6 digits before the decimal point". The input now lowers the ceiling to whatever the field's own precision can express, so F 1.1 clamps at `999999.99` while the F2 and F3 fields keep the full value.

